PR #69, opened 2021-05-12
Dynamic resolution support

ds-sloth:
I have been impressed with how actively your development has proceeded over the past few months and would like my 3DS port to benefit from ongoing development in the future.

In order to do so, I want to merge my changes back into master. I am going to issue a series of pull requests to merge these changes; this is the first one.

In this pull request, I include most of the changed drawing code from the 3DS port. The major change facing PC users is non-800x600 resolution support. I support resolutions from 512x384 (the DS's resolution) to 1280x720. Right now 1280x720 is not completely ready due to incompatibilities with Backgrounds 1, 2, 13, 17, 22, 35, 36, 37, 51, 52, and 56, so I would not advise you to merge immediately. There are some assets that I have created that make 1280x720 look better. I don't know how you would like me to contribute those.

I am aware of the range of changes that dynamic resolution can cause because I have found them while playtesting my 3DS port. The biggest workaround I have added is for events. A lot of episodes have events that assume that the resolution is 800x600, so any time the controls are locked, NPC activation events are calculated in 800x600.

In addition, this code includes changes to support a full editor and has also refactored UpdateGraphics to include all logic code at the top of the function, instead of having two copies of the logic code, one in the frameskip section and one in the main function.

I will add support for the Backgrounds 1, 2, 13, 17, 22, 35, 36, 37, 51, 52, and 56 in a commit tomorrow and would request you to merge this PR at that time.


ds-sloth:
I have added the Background fixes (including some unexpected horizontal fixes) so these three commits are ready to be merged into the main branch if you are interested. Thank you for your consideration!

ds-sloth:
I just added vertical looping for Background2s in extremely large areas. Definitely see why you want to fix Background2 handling in general.

Wohlstand:
Folks at my Discord server were tweaking and actually reviewing the thing, however, I didn't review this yet as I feeling not good by myself, and I need to get some more energy until I will be able to review the whole thing by myself. However, if you want to participate discussion, feel free to join my Discord server :fox_face: :+1:

Wohlstand:
btw, can you try to loop the curtain sprite at the top of the screen? Otherwise, folks getting something like this (picture below), or you had to fix this already?
изображение

ds-sloth:
Let's not rush to merge this if you are not currently making rapid changes. I will join your Discord. And yes!
Interface4
MenuGFX1

Wohlstand:
Another notice from the chat that flying cheap-cheeps on a widescreen getting jump earlier than needed. I guess, they should jump in only the specific distance closer to the player than the "on-screen" appearence.

Wohlstand:
The fun fact: when I made a little test at PGE Engine with the "4k SMBX", I accidentally caused the section change event being triggered earlier than needed, caused my playable character being sucked into the lava hole directly :grin:

Wohlstand:
BTW, This is my very old experiment I did two years ago, and I had to re-make it again to make the video:
https://user-images.githubusercontent.com/6751442/118053245-d711e680-b38c-11eb-9d29-7a3643d24916.mp4

The section resizing event gets activated earlier than needed, and in result, player gets trapped at the hot oiled hole

Wohlstand:
Another case folks do report: the background should be aligned at the actual edge of the section, in result, this:
изображение

Wohlstand:
Also, please make sure all out-of-section spacings gets always filled, especially when the screen shaking got proceeded by any reason: POW-block, twomps, clown-car, Bowser IIIrd, etc.

ds-sloth:
(The most recent commit is for NoTurnBack but I accidentally misformatted the commit message.)

src/editor.cpp
```diff
@@ -2056,22 +2060,23 @@ void GetEditorControls()
...
+        // else
+        //     EditorControls.Down = false;
+        // if((getKeyState(vbKeyLeft) == KEY_PRESSED))
+        //     EditorControls.Left = true;
+        // else
+        //     EditorControls.Left = false;
+        // if((getKeyState(vbKeyRight) == KEY_PRESSED))
+        //     EditorControls.Right = true;
+        // else
+        //     EditorControls.Right = false;
```
Wohlstand, line 2079: Are you sure you didn't break anything?

src/main/menu_main.cpp
```diff
@@ -331,21 +368,23 @@ bool mainMenuUpdate()
...
         if(SDL_AtomicGet(&loading))
         {
             if((menuDoPress && MenuCursorCanMove) || MenuMouseClick)
                 PlaySoundMenu(SFX_BlockHit);
             if(MenuCursor != 0)
                 MenuCursor = 0;
         }
+#endif
         // Main Menu
         else if(MenuMode == MENU_MAIN)
```
Wohlstand, line 381: Haha, you disabling the block, and missing up the `else if`, that will break the compilation on 3DS itself :grin:

in such case, you should take `else` into the macro too:
```cpp
#ifndef __3DS__
    if(...)
    {
    }
    else
#endif
    if(...)
    {
    }
```

src/main/menu_main.cpp
```diff
@@ -421,7 +460,7 @@ bool mainMenuUpdate()
                     MenuMode = MENU_BATTLE_MODE;
                     menuPlayersNum = 2;
                     menuBattleMode = true;
-#ifdef __EMSCRIPTEN__
+#if defined(__EMSCRIPTEN__) || defined(__3DS__)
```
Wohlstand, line 463: I would simplify this:
- at the top of the file, add next:
```cpp
#if defined(__EMSCRIPTEN__) || defined(__3DS__)
#define MENU_NO_THREADED_LOAD
#endif
```

- in the code, simply use `#ifdef MENU_NO_THREADED_LOAD` and `#ifndef MENU_NO_THREADED_LOAD`

src/main/menu_main.cpp
```diff
@@ -1532,11 +1533,10 @@ void mainMenuDraw()
             SuperPrint(fmt::format_ne("{0} {1}%", g_mainMenu.loading, progress), 3, MenuX, MenuY);
         }
     }
-#else
-    if(false) {}
+    else
 #endif
     // Main menu
-    else if(MenuMode == MENU_MAIN)
+    if(MenuMode == MENU_MAIN)
```
Wohlstand, line 1539: When the "loading" flag is active, you should keep `else`, to don't cause the menu to be drawn, here you would get a crash because, in a moment of loading, the menu gets filled in a separated thread.

Wohlstand: Okay, I checked the code, now it's fine :fox_face: :+1: 
sorry

src/editor.cpp
ds-sloth: Definitely broke the editor controls for now. Just want to move all of the EditorControls logic into a single place closer to `joystick.cpp` and not in `editor.cpp`. Which is why I wrote TODO.

src/main/menu_main.cpp
ds-sloth: Fixed now, thanks!

src/graphics/gfx_update.cpp
```diff
@@ -664,17 +673,16 @@ void UpdateGraphics(bool skipRepaint)
...
                     if(!NPC[A].Active)
                         onscreen = false;
                 }
 
                 if(can_activate)
                 {
-                    if(NPC[A].Generator)
-                        NPC[A].GeneratorActive = true;
-                    else if(NPC[A].Type == 0) // what is this?
+                    if(NPC[A].Type == 0) // what is this?
```
Wohlstand, line 685: This is a buffer NPC used for Editor or magic hand NPC rendering

src/graphics/gfx_background.cpp
```diff
@@ -46,7 +46,7 @@ void DrawBackground(int S, int Z)
     A = 1; // Blocks
     if(Background2[S] == 1)
     {
-        int tempVar = static_cast<int>(floor(static_cast<double>((level[S].Width - level[S].X) / GFXBackground2Width[A] * 0.5))) + 1;
+        int tempVar = static_cast<int>(floor(static_cast<double>((level[S].Width - level[S].X) / GFXBackground2Width[A] * 0.5 + (double)ScreenW / GFXBackground2Width[A]))) + 1;
```
Wohlstand, line 49: btw, this monstrous `static_cast<int>(floor(static_cast<double>(....))` is the result of the VB to C++ mess, you may replace them with just an int cast.

src/frm_main.cpp
```diff
@@ -99,8 +99,10 @@ bool FrmMain::initSDL(const CmdLineSetup_t &setup)
     //Write into log the application start event
     pLogDebug("<Application started>");
 
+#ifndef NO_SCREENSHOT
     m_screenshotPath = AppPathManager::screenshotsDir();
     m_gifRecordPath = AppPathManager::gifRecordsDir();
+#endif
```
Wohlstand, line 105: Yeah, on Emscripten they are fully useless as no way to download them developed right now

ds-sloth:
These changes will be merged at a later time using the `wip-multires` branch.

PR #119, opened 2021-05-28
Improving readme spanish translation

M0REKZ:
I fixed some things caused probably by google translator, now the text is more easy to understand, the only word hard for me to translate is "warp", maybe someone else can choose a word in Spanish for that... but maybe it's unnecessary


Wohlstand:
Gracis :fox_face: :+1:

M0REKZ:
ok, I will do that in future spanish translations for other things

Issue #72, opened 2021-05-16
Bug: Unfrozen Bullet Bills Remain Stationary


ChristianSilvermoon:
If you freeze a Bullet Bill using an Ice Flower, and then Unfreeze it using a Fire Flower, it remains stationary.
![Scr_2021-05-16_12-23-29](https://user-images.githubusercontent.com/30162808/118410291-79f38a80-b643-11eb-8475-634d3c125bc9.gif)

GIF is from [ds-sloth's repo @ 92d12fd](https://github.com/ds-sloth/TheXTech/commit/92d12fd4abbfa9e8edbb7bd64132275f0c382c45)

but issue appears to originate from here

Wohlstand:
Yeah, I can confirm:
![Scr_2021-02-10_01-32-50](https://user-images.githubusercontent.com/6751442/118413803-cd98bf00-b6a9-11eb-950e-4eee831a5427.gif)

It's my bug, at the vanilla game it works fine:
![Peek 2021-05-16 22-33](https://user-images.githubusercontent.com/6751442/118413820-f0c36e80-b6a9-11eb-8ab7-b6ab17b81947.gif)

PatriotRossii:
I wanna try to dive into this and fix it

ds-sloth:
The truth is even stranger than expected.
The direction of the bullet was being passed by reference to `EditorNPCFrame` when it was unfrozen, and `EditorNPCFrame` randomly sets the direction when it is zero.
I have updated `EditorNPCFrame` to take its direction parameter by reference.
I assume that was not what you expected, so I have also added a compat setting (on by default) that makes the bullet bill cannon set the bullet bill's `DefaultDirection` to its `Direction`, making it always keep its direction when it is melted.

Vanilla behavior:
![Scr_2021-06-02_01-27-14](https://user-images.githubusercontent.com/72112344/120428824-5a967600-c342-11eb-81e1-e32b6f764322.gif)

New default behavior:
![Scr_2021-06-02_01-28-12](https://user-images.githubusercontent.com/72112344/120428846-62eeb100-c342-11eb-8912-bcc1e28f507b.gif)

Wohlstand:
I had to test this, confirmed it's fixed! :fox_face: 
Thank you very much for the fix! :fox_face: :+1:

PR #128, opened 2021-06-02
Bullet bill fix (EditorNPCFrame -> pass direction by reference)

ds-sloth:
This fixes the bullet bill unfreezing issue #72 by bringing `EditorNPCFrame` in line with its vanilla behavior, which is to pass the NPC's direction by reference. This is a significant change, but it is how the original game functions.
Additionally, it adds a compat option to get the expected behavior, which is for cannon-fired bullet bills to keep their original direction after being frozen and thawed.


Wohlstand:
Wow, it's just the SAME bug as I had at the 1.3.0 version against points multiplier as it gets value passed as a reference. Thank you very much for the fix!

PR #150, opened 2021-07-02
Add multires branch on-push CI

Nomagno:


Wohlstand:
You better push that into the branch you need, here this won't work obviously

PR #155, opened 2021-07-14
[NEW BRANCH] Wip vita support

suicvne:
Merging changes I've made from suicvne/TheXTech@wip-vita-support into the new branch wip-vita-support in Wohlstand/TheXTech.

NUMEROUS configuration file changes and more! This is all assuming you have the vitasdk installed on your machine.

Known quirks
1. SDL Video mode doesn't work at all, so an alternative renderer based on VitaGL is provided.
2. Large Background images are currently NOT being loaded in, this is temporary as I sort out memory usage.
3. AudioCodecs are not built, instead SDL-mixer-x is built as a static library against the VitaSDK's pre-patched audio codecs. 
4. Not all audio codecs are supported
5. Sound doesn't work anyway with SDL_mixer (something about improper format when trying to start) 
6. FreeImage/GraphicsHelps don't seem to work, unsure why yet. Vita works around this by using stb_image.h and stb_image_resize.h.
7. stb_image_resize.h has bugs on certain image types. The real fix is to get FreeImage/GraphicsHelps:: working again.

Known configuration quirks
1. -DVITA doesn't seem to always be passed
2. Doesnt build with CMAKE_RELEASE_TYPE=Debug (SDL_mixer_x builds with d at the end, causes linking errors). (THIS NEEDS to be fixed, as core dumps are useless without building in Debug mode and getting -g -O0)
3. -O3 is passed anyway because of no debug mode. 

Configure:
```
git clone https://github.com/Wohlstand/TheXTech.git
git checkout wip-vita-support
git submodule init
git submodule update

mkdir build
cd build
cmake -DCMAKE_TOOLCHAIN_FILE=${VITASDK}/share/vita.toolchain.cmake -DVita=1 -DVITA=1 
```

This version can be debugged remotely with `socat` see `lib/Logger/private/logger_vita.cpp` for more information.


suicvne:
Wohlstand granted permission for me to merge these, so long as they're on their own branch.

Issue #171, opened 2021-08-29
Battle levels list is empty on master branch


Wohlstand:
Recently, Derpy was found a bug: the battle levels list is empty at the `master` branch
![изображение](https://user-images.githubusercontent.com/6751442/131264213-10e9a48d-a56c-4121-a9e9-e69401a3ec6d.png)

I do confirm the same:
![изображение](https://user-images.githubusercontent.com/6751442/131264221-e3728438-9437-4b89-ab17-24581399b212.png)

At the `stable-1.3.5.x` branch and at older versions everything works fine.

This bug has come with an attempt with an effort of @ds-sloth to separate the battle levels list and pre-load all lists before the game will complete its loading, unfortunately, there is a defect that causes the battle levels list to be completely empty. I didn't investigate for details of this bug yet, but I bet it's simple to fix, however, this is a serious bug that needs to be fixed.

Wohlstand:
Bug has been fixed by @suicvne, confirm it works fine now :fox_face: :+1:

PR #173, opened 2021-08-29
Fix Issue #171:

suicvne:
- Use proper SelectBattle vectors to load battle levels from (game_main.cpp and menu_main.cpp) 
- DONT call FindSaves() if the MenuMode is MENU_BATTLE_MODE to prevent segmentation fault when selecting a level.

I have confirmed that these fixes allow me load and play Battle Mode levels on Linux. I have not yet tested on Vita. These changes were relatively minor so I have no reason to believe that this should not work on Vita. However, I am going to test anyway.

image

### DIFF
--- a/src/game_main.cpp
+++ b/src/game_main.cpp
@@ -1552,7 +1552,7 @@ void StartBattleMode()
     PGE_Delay(500);
     ClearLevel();
 
-    if(NumSelectWorld <= 1)
+    if(NumSelectBattle <= 1)
     {
         MessageText = "Can't start battle because of no levels available";
         PauseGame(1);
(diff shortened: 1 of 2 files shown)


Wohlstand (approved):
For me looks fine, gotta test that on my phone once Android Ci will build the thing. I do see the Ci failed at your vita branch, but seems the issue is different one.

suicvne:
Yes, the Vita build failing was because the vitaGL update finally went live and I needed to add a new library to the linked library list. 
Vita build is passing now and I was able to test the changes and load a battle level:
hypothetical_2
hypothetical_3

The line is borked for the same reason some rects are borked, but I plan to look into that still separately

PR #191, opened 2021-09-22
Devel mixer fixes

Wohlstand:
I made some check out and I made my own fixes on this:
- Please keep the code with 4 spaces than tabs
- Keep call sigantures same (they should return int to indicate the possible error: 0 the fne, and -1 an error)
- If call "didn't properly declared", use the same macro as in original header, but instead of Mixer'x call, pass any your equivalent call

### DIFF
--- a/src/3ds/mixer_platform.cpp
+++ b/src/3ds/mixer_platform.cpp
@@ -60,7 +60,7 @@ bool MixPlatform_NoPreload(const char* path)
         return false;
 }
 
-void MixPlatform_PlayStream(int channel, const char* path, int loops)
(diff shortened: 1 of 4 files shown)


ds-sloth:
Thank you Wohlstand! All of this looks very good.

I am sorry about those tabs. I dislike them too, but my Sublime text settings got reset and I have been writing tabs by default without wanting to.

So, the only thing I am unsure of is why we define the ticks parameter for the play channel given that we don't use it in TheXTech and it would take time to implement on the current 3DS lib.

I'll check that it builds on 3DS but you can merge it in the devel first since I think I am the only one making 3DS builds right now.

Issue #156, opened 2021-07-16
Feature Request: Display a Level's Star Count on the World Map


0lhi:
Doors and Warps to other Levels can display their Star Count. It would be nice to bring this feature to Worldmaps as well. (Particularly for Episodes that have no [Star Tracking Levels](https://i.postimg.cc/zGhZfBxL/2020-03-19-01-06-51.png)).

![StarCount1](https://user-images.githubusercontent.com/74732674/125937532-56f9aeec-6386-44ad-b41f-1070bd93dbe1.png)

Dragon Coins, too, once #122 is done.

![StarCount2](https://user-images.githubusercontent.com/74732674/125937591-44c91b4a-974a-4f99-8383-7f3f5d3dc403.png)

Wohlstand:
Stars count on the world map seems a good thing, that's really easy to do, and so:
- Levels that has no stars, will show nothing
- Add WLDX-side flags:
-- map-wide: enable/disable/unset stars showing
-- level-point local: enable/disable/unset stars for this point only
- `compat.ini` will define the default value for unset values

Wohlstand:
I guess, this thing should majorly simplify the episodes playthrough that has hidden stars on certain levels and it's hard to figure them, but these stars were required to complete the episode

Wohlstand:
Okay, this task has been completed. Any other additional things will go as separate tasks.

Issue #37, opened 2021-02-09
Game Over screen


Wohlstand:
Since everything, this game hadn't the proper game over screen, and there are next ideas:
- **Simple:** Show the "Game Over" on a black screen with the playing of the specific music, then, show the question to continue, or quit into the main menu.
- **More tricky:** In addition to simple, play some animation effects until the continue/quit question.
- **Like 38A does:** When a player has lost all lives, start the level specified as a "game over level" at the world map. When this level gets completed with any result, quit to the main menu.
- It's possible to don't show the question, just quit to the main menu, depending on settings.
- Optionally, add to the world map an option to disable the lives counter making them unlimited and useless (suggested for kaizo and other tricky or just hard episodes) (and let life shrooms act differently in this condition).

MarioLover63:
Is it too much to ask for all three?

Wohlstand:
> Is it too much to ask for all three?

It's possible to make work all of them depending on a world map settings and use the first one as the default for everything.

ChristianSilvermoon:
I like the idea of all three

But I also like the idea of a generic "Game Over" level being used as a background for a Game Over Screen Menu thing, much like the Title and Credits Levels... maybe something like that could be overridden per world?

0lhi:
> Optionally, add to the world map an option to disable the lives counter making them unlimited and useless (suggested for kaizo and other tricky or just hard episodes) (and let life shrooms act differently in this condition).

Regarding the difficulty mode that we talked about, I looked at what other games are doing, and 4 approaches came to mind:

**Vanilla:** Start with 3 1UPs, get up to 99.

**Nuanced:** Start with 5 1UPs, get up to 999. (Same as SMA4, max resembles 38A, start resembles NSMB & All-Stars)

_**Dynamic:** 3 1UPs at the beginning, increase the amount after every game over, get up to 9999. (Resembles Sonic Robo Blast 2)_

**Casual:** Infinite Lives. (Resembles Odyssey, Crash 4 & Rayman Redemption)

In the case of Rayman Redemption's casual mode, the life counter is set to ∞. 1UPs turn into Tings (≈ Coins), and the Ting Counter doesn't stop at 99.

In Super Mario Odyssey, when a Player dies, he loses some of the Coins he collected.

You could combine these approaches to have something like this:

![Casual](https://i.imgur.com/olsh1i1.png)

[...]

PR #192, opened 2021-09-26
Show collected stars per level at world map

Wohlstand:
All changes was done without WLDX new fields yet, will add them later


ds-sloth:
I'm not sure that this should be the default behavior.

In many episodes (like Invasion 2) it is unknown which levels have stars. I would by default only print the number of obtained stars, not the number of available stars. There could be a wldx setting that allows to print number of available stars. How does that sound to you?

0lhi:
Shared my Take in the Moondust Chatroom, gonna paste here:

>Oh no, I strongly think that it should be turned on by default :o It is the case for Hubworld Episodes already, and no-one sees Spoilers as a concern there
>
>Plenty of people that I introduced to The Invasion 2 were turned away by it's confusion Map Layout, and ended up staying away from SMBX as a consequence
>
>The Star Locations being hidden does contribute to that confusion
>
>It's why I currently recommend people new to SMBX to start with "Super RMN World", because there it's clear how many stars are where
>
>People who want to get the Vanilla Experience can just turn to Mode 2 or 3. Afterall, that's what they are there for
>
>But TheXTech as a whole shouldn't suffer from the limitations of Vanilla SMBX. Lack of Star Display is rarely an intentional decision, but almost always a consequence of it's limitations.

ds-sloth:
From the chat it seems that everyone wants to have the number of acquired stars shown by default and the question is whether the number of available stars should be shown by default. It's clear from our level of disagreement that this should be a settable config option (show or don't show star locations) with both available for the modern game and the question is which setting should be on by default for the modern game.

My philosophy behind modern mode is that it should be the best way to experience the gameplay of SMBX, with modifications that make it look better and feel more modern or provide the player with information they already have, but without altering the overall experience of the game. I see showing the star locations as a core experience modification, similar to adding spin jumps to characters or adding ground pound without the correct yoshi.

If the star locations in an episode are too confusing, that might be a problem with the episode, but the creator of the episode made the decision to hide the stars (as they are hidden in many of the games that inspired SMBX and TheXTech). I do think that at least in the games that inspired SMBX, not displaying star locations (or warp cannon locations, later on) was completely intentional.

From my perspective, a lot of the excitement of episodes like The Invasion 2 is in finding the hidden stars. I don't think that effectively adding a walkthrough by default would improve most users' experience, especially given that it's almost never required for a user to find any stars in the episodes where they are hidden. (And if it is, that's a major design flaw with the episode but not our job to fix by default.)

I propose that we should set up some sort of poll to figure out the default. Should we do that on GitHub or on Discord? (Preferably both, but I'm not sure how to merge them.)

Wohlstand:
Okay, I had to extend the functionliaty of this mechanism, and now there are many ways to set up the showing of colelcted stars:
- per level individually
- per whole world map
- per compat.ini
- and the last - by settings/thextech.ini

0lhi:
> My philosophy behind modern mode is that it should be the best way to experience the gameplay of SMBX [...] but without altering the overall experience of the game.

I see a contradiction there. The Vanilla Engine has many flaws, so the best experience entails Changes. It's why many Compat-Fixes are enabled by default. This would just be another one on the list.

> If the star locations in an episode are too confusing, that might be a problem with the episode, but the creator of the episode made the decision to hide the stars

In most cases I think it was just not caring to make a Star Tracking Level (which is bothersome) rather than the active decision to hide. The Original Issue #156 I made was because of the NSMB DS Remake, which is an example for that. In The Invasion 2, you can already check the Map's Star Entrances by entering the doors in Rosalina's Place, so for this episode, the Change seems more like a Quality-of-Life thing to me, similar to *enable-last-warp-hub-resume*, rather than a Spoiler.

> I do think that at least in the games that inspired SMBX, not displaying star locations (or warp cannon locations, later on) was completely intentional.

Umm, can you name some examples? Warp Zones are one thing. There weren't any Stars in the old 2D Marios, but the Map did share which levels contained Secret Exits. Mario 64 does share most of the Star Locations. Secret Stars are an exception, but in the case of those, the levels themselves are hidden. So, the Analogy for SMBX would be revealing the Map Path before unlocking it, which I do not want.

> From my perspective, a lot of the excitement of episodes like The Invasion 2 is in finding the hidden stars. I don't think that effectively adding a walkthrough by default would improve most users' experience

Walkthrough is too big a word. The Secret Exits are announced on the Map, but it's still up to you to find them. Same would be the case for the Stars. All it does is removing the guess work on which level contains how many.

> (And if it is, that's a major design flaw with the episode but not our job to fix by default.)

I mean, nothing we do here is our Job. We contribute voluntarily to improve the Playing Experience. This change in particular has the potential to turn countless of bad episodes into good episodes by changing one single detail. That seems like a pretty good deal to me.

> I propose that we should set up some sort of poll to figure out the default. Should we do that on GitHub or on Discord? (Preferably both, but I'm not sure how to merge them.)

You mentioned your Philosophy above. My philosophy / source of motivation is to make SMBX as accessible and inviting as possible. I think an intuitive experience is critical for that, and my observations so far have confirmed this.

A Poll that matches the preference of people who are already inside the SMBX Community may not represent what would make TheXTech appealing to New Users. So, I think the default should be decided by debate rather than a vote.

ds-sloth:
> A Poll that matches the preference of people who are already inside the SMBX Community may not represent what would make TheXTech appealing to New Users. So, I think the default should be decided by debate rather than a vote.

I can get behind this. I similarly prefer to discuss until consensus is reached.

> I see a contradiction there. The Vanilla Engine has many flaws, so the best experience entails Changes. It's why many Compat-Fixes are enabled by default. This would just be another one on the list.

I see a continuum between (1) things like fixing bugs and changing the level resolution and (2) altering the rules of the game / the way a player approaches the game. In our past discussions, I have often been more hesitant to change the gameplay (such as preferring not to allow repeated drop/add without costing lives) than some of the other community members.

One option to resolve these competing philosophies would be to offer users, upon first boot, three experiences -- "modern gameplay", "preserved gameplay", "preserved everything" -- with corresponding sets of default compat values and gameplay settings. I don't know how that would strike the other people in this discussion.

Wohlstand:
Note: default global behavior is: showing is disabled, user will be able to enable this by settings/thextech.ini or by compat.ini, or later by WLDX (when I will implement Editor-side fields binding, `$%@&` my 2014's code...)

0lhi:
> One option to resolve these competing philosophies would be to offer users, upon first boot, three experiences -- "modern gameplay", "preserved gameplay", "preserved everything" -- with corresponding sets of default compat values and gameplay settings. I don't know how that would strike the other people in this discussion.

Great idea! Who are we to decide which approach is the correct one? Letting the user choose at first boot is the way to go 👍

It reminds me a bit of the [Difficulty Mode](https://github.com/Wohlstand/TheXTech/issues/37#issuecomment-781045111) that I brought up in Wohl's Game Over thread. Maybe the approaches could be combined somehow?

I think, the question on how to do this in detail warrants a new GitHub Discussion. But it's definitely the direction we should take.

ds-sloth:
> It reminds me a bit of the Difficulty Mode that I brought up in Wohl's Game Over thread. Maybe the approaches could be combined somehow?
I agree wholeheartedly. Start 5, max 999 lives (or even casual mode?) seems like a clear default for the "modern gameplay" version. (Casual mode makes kaizo episodes actually somewhat reasonable... :) )

In any case, @Wohlstand, I approve of the changes you proposed in this PR. Let's move discussion of default options to a GitHub Discussion for the future, since it seems that the discussion is becoming increasingly broad in scope.

Wohlstand:
> In any case, @Wohlstand, I approve of the changes you proposed in this PR. Let's move discussion of default options to a GitHub Discussion for the future, since it seems that the discussion is becoming increasingly broad in scope.

Right, so, gonna merge this. :fox_face: :+1:

Issue #197, opened 2021-10-02
Suggestion: A new version of the has-item (sound-12) sound


vitzie629:
In Super Mario World and Super Mario All Stars, when the player already has a power-up and touches the same power-up as earlier (example: Fire Mario in SMB, SMB3, and SMW touches a fire flower), the player-grow sound plays for SMB and SMB3, and for SMW, the player-grow and has-item sound plays together. The old has-item sound should be replaced with a new one which is the has-item and player-grow sounds combined together to make it sound like SMW.

[New has-item sound.zip](https://github.com/Wohlstand/TheXTech/files/7270678/New.has-item.sound.zip)

Wohlstand:
The fact I don't need to replace the sound actually, but I can play two sounds in parallel :)

vitzie629:
> The fact I don't need to replace the sound actually, but I can play two sounds in parallel :)

How do you play two sounds in parallel?

Wohlstand:
> How do you play two sounds in parallel?

Easily, in the code making two lines of the sound play:
```cpp
PlaySound(SFX_PlayerGrow);
PlaySound(SFX_GotItem);
```

vitzie629:
> > How do you play two sounds in parallel?
> 
> Easily, in the code making two lines of the sound play:
> 
> ```c++
> PlaySound(SFX_PlayerGrow);
> PlaySound(SFX_GotItem);
> ```
> 
> This call is fully asynchronious, once I call, the sound starts to play, calling them two, the two sounds will be played

Thanks! Will you do that for TheXTech 1.3.5.2 or 1.3.6?

Wohlstand:
I made a quick test, I madded this behavir into hammer suit taking (kept old sound behavior for leaft and other things), so, see the result:
https://user-images.githubusercontent.com/6751442/135699059-26773d70-293d-419d-a32f-b41ea9077740.mp4

vitzie629:
> I made a quick test, I madded this behavir into hammer suit taking (kept old sound behavior for leaft and other things), so, see the result:
> https://user-images.githubusercontent.com/6751442/135699059-26773d70-293d-419d-a32f-b41ea9077740.mp4

I saw the result and the hammer suit test with the two sounds together sound great!

0lhi:
I'm not sure where's the logic behind playing the grow sound when the Character isn't growing.

This should be optional and opt-in.

P-Star7:
[...]

Issue #185, opened 2021-09-10
SMBX Pre-1.3 compatibility


Wohlstand:
To make proper work of older levels and episodes developed for SMBX versions older than 1.3, we need to provide the compatibility for given file versions and set up different behavior for features. Some of them were done already.

The full list of pre-1.3 compatibility features:
- [x] Different behavior for Bowser the Third
- [x] Billy gun (NPC-22) should shoot 2x time slower when version less than 28 and should keep shooting while projectile if version less than 51.
- [x] Thwomp (NPC-37) should always fall with no matter to the Y coordinate of a player. Version less than 9.
- [x] Need to fix bug #169 with the platform blocks (NPC-60), right now here is a workaround made. Otherwise, the platform will be stuck at the "Clown car" level of "The Invasion" episode. Version less than 9.
- [ ] Disable the spin-jump ability for players at builds before 20
- [ ] Make Luigi physics act the same as Mario at builds before 51
- [ ] Before build 51, players shouldn't be able to grab a shell from above
- [ ] Toad shouldn't be able to spin-jump when a version is older than 64.
- [ ] Boo (NPC-38) should fly a bit faster
- [ ] Mario and Luigi should automatically drop the reserve item when the version is older than 51
- [ ] Before X version (probably version 30), have the higher section top limit (that's required to get the secret exit at "P-Switch Panic" level because the flying over the wall is required) [NEED AN INFO]
- [x] Before build 58, Mario and Luigi only existed (#223)
- [ ] Before build 51, you can't choose between Mario and Luigi as a first player (#223)
- [ ] Before build 30, Thwomps, Boos, and Eeries are fully immortal, nothing can kill them

## Notes
- All pre-1.3 compatibility features should work when the compatibility level is set to MODERN (Mode 1). If the compatibility level is SMBX2 (Mode 2) or SMBX1.3 (Mode 3), all pre-1.3 compatibility features should be disabled.

ds-sloth:
This looks really exciting to me and should be a great way to preserve SMBX history.

How can we determine the exact speed of NPC-38A in older versions? Do we need to reverse engineer? (Too bad we only have the V64 source and not each version's source.)

LoveBodhi:
Toad can spin jump since 1.3

Wohlstand:
>  NPC-38A -> NPC-38

Damn, by reflexes I type A after I type "38" :rofl:

Wohlstand:
[...]

Issue #163, opened 2021-07-23
Suggestion: Target Versions and Backward-Compatibility for lvlx/wldx


0lhi:
One flaw of SMBX2 is that it isn't backward-compatible to it's previous versions. Episodes made for Beta 3 tend to be incompatible with Beta 4. So, if one wants to ensure that an episode will remain compatible with future versions, it essentially has to be made for 1.3.

This is true for TheXTech as well. Mode 3 will stay the same, but everything else is subject to change.

For Vanilla Episodes, the goal is to implement the #54 MD5 system that enables whatever compat flags are necessary. But I don't think that this effort should extent to XTech-targeted episodes. At some point it would be too much to keep track of. So instead I'd like to propose a different solution:

**1. Add a** `[backward-compatibility]` **section to Compat.ini for TheXTech's bugs and peculiarities.**

This doesn't have to be done retroactively, since there are no built-from-the-ground-up episodes yet (that I'm aware of), but it would be a good preemptive measure. The #72 Bullet Bill compat setting made by @ds-sloth could be moved there.

**2. In the future, keep track of the default compatibility settings for every release version**

Taking both, Vanilla and TheXTech options, into consideration.

**3. Allow to specify in the wldx/lvlx file which release version of TheXTech is targeted.**

If it's targeted for *1.3.6.1*, *1.3.6.2* would simulate it's behavior. *1.3.6* would give a warning message that the episode is intended for a higher version, and allow the user to choose between:

- Playing at the risk of bugs occurring
- Opening a link to the newest version.

The hierarchy would be: Level Compat.ini > Episode Compat.ini > Target Version default settings > Global Compat.ini

Wohlstand:
Right now every LVLX/WLDX file contains the config pack key name that was made to prevent possible faults. Anyway, there is reason to take the version identifier from the assigned TheXTech binary (make certain embedded string at the executable that can be easily recognized by the file scanning) to apply it to the current working files, but give the ability for users to choose the compatibility settings. (probably the good thing to add the dirtily-made compat.ini editor that supports all supported options, but later)

0lhi:
TheXTech has many goals, many of which are in competition with each other. Unfortunately backwards-compatibility isn't feasible without imposing too severe restrictions elsewhere, so this idea didn't make the cut.

[...]

PR #198, opened 2021-10-02
Play both player grow and got item sounds

Wohlstand:
To match behavior of original games
#197


ds-sloth:
As I mentioned in Discord, this is a cool feature but should probably be configurable user-side. Let's protect this new code by some `g_config` option for now and then merge it.

0lhi:
> As I mentioned in Discord, this is a cool feature but should probably be configurable user-side.

Agree with that, and it should be disabled by default.

Wohlstand:
I updated some, could you review the thing? I added thextech.ini option and I made compat.ini option that forces to disable the feature for episode/level.

Wohlstand:
Note, last 3 commits I posted here accidentally, I cherry-picked them into the master branch immediately

0lhi:
Disabling/enabling the ```enable-player-grow-with-got-item = 0``` setting works. The compat.ini succesfully overrides the enabled setting.

**It should be disabled by default**, however. Reason for this is that the combination of those particular sound effects can lead to a sensory overload for some people. It does for me, and it's not pleasant.

This may have been the original setting in the original SMW game, but there is a major difference between it and SMBX: There are no Power-Up generators. You are much less likely to encounter that soundeffect combo in the SNES game, while in SMBX, you may end up hearing it several times per second. Judging by this [TIP](https://www.smwcentral.net/?p=viewthread&t=59235) from SMWCentral:

Items

Hacks also seem mostly unaffected by it.

Players who like the SMW approach will probably notice the difference and can just toggle the setting. The opposite is less likely to happen, since it's not immediately obvious there is an alternative available.

Wohlstand:
@0lhi, made a fix. Also, I added the compat.ini side equivalent of the global settings to have an ability to enable that sound behavior even global settings disables it (I'll add this into the side of "Return to Dinosaur land" copy I have)

0lhi:
> Also, I added the compat.ini side equivalent of the global settings to have an ability to enable that sound behavior even global settings disables it

That would have the issue that it imposes a setting on the players that goes against their individual preference.

This option doesn't affect gameplay in any way, so I don't think there should be a compat feature. Just let the players choose what they like. Everything else would needlessly complicate things.

Wohlstand:
At least, this option is for episode makers who, for example, make SMW-themed episodes, and want to make them act more authentic, so, adding "ignore" setting and own value at the side of the episode, or oppositely, forced to disable it as while this option enabled, it may cause some issues. Just an option. This has a more critical case when sounds were been customized into something unusual that makes a crap, so, that compat.ini option also should help to prevent the future crap caused by an unexpected combination of sounds.

0lhi:
Optional authenticity is good, mandatory in this case is not. Like I said above, it can cause irritation for some people, particularly for [HSPs](https://en.wikipedia.org/wiki/Sensory_processing_sensitivity). To what extend depends on the individual, but speaking for myself, it makes me cringe. I don't like it.

> that compat.ini option also should help to prevent the future crap caused by an unexpected combination of sounds

This I agree with. I guess, the best option would be to offer these global settings:

```ini
[effects]
player-grow-with-got-item = enable | 2 (activate unless compat.ini forces disable)
player-grow-with-got-item = allow | 1 (disable unless compat.ini says to enable)
player-grow-with-got-item = disable | 0 (force disable, default)
```

Wohlstand:
> EDIT: Also, compat.ini should probably get more categories. This sfx thing doesn't really have to be under [compatibility] since it's not about gameplay.

There is already for speedruns. I should add the "[effects]" section then

Wohlstand:
So, just now I updated the thing, and now:

settings/thextech.ini:
```ini
[effects]
player-grow-with-got-item = disable; | 0 (force disable, default)
player-grow-with-got-item = allow; | 1 (disable unless compat.ini says to enable)
player-grow-with-got-item = enable; | 2 (activate unless compat.ini forces disable)
```

compat.ini:
```ini
[effects]
ignore-got-item-setting = false/true ; seize the control over global setting value
enable-player-grow-with-got-item = false/true ; set the actual value for this setting
```

ds-sloth:
I'm confused here; I would think that the `compat.ini` should be able to override even the main `disable` setting.
Additionally, I think that `compat.ini` should use a single setting with values `allow`, `force-on`, and `force-off`, instead of two settings. If you want, you could even have a single setting with `suggest-on`, `suggest-off`, `force-on`, and `force-off`.

Wohlstand:
I guess that's more reasonable. I really thought to make next scheme:
- global set to disabled by default
- when episode-side defines force-on, the force-on should work with no matter what global sets, and the sort of "unspecified" (by default) that means global settings will define the behavior

I'll try to make an update tomorrow, but keeping discussion up until we find the best solution which is fine for all of us.

> Семь раз отмерь, один раз отрежь!

0lhi:
> I would think that the compat.ini should be able to override even the main disable setting.

@ds-sloth I disagree for the reasons I mentioned above. Imposing the setting on players would make for an unpleasant experience for those who have above-average sensitivity to sounds.

I don't mind if the author can force-disable it. But the player **must** be able to force-disable it also.

> Additionally, I think that compat.ini should use a single setting [...], instead of two settings.

Agree there should just be one setting. I think that could be done by just removing the ```ignore-got-item-setting``` setting.

* If ```sfx-player-grow-with-got-item``` at settings.ini is set to ```disable``` (default), it would override the compat.ini value.

* The compat.ini value ```true``` would override the settings.ini value ```allow```.

* The compat.ini value ```false``` would override the setting.ini value ```enable```.

This adds plenty of room for customization, allows to prevent the unexpected combination of two custom soundeffects, and avoids imposing a potentially irritating arrangement on HSPs such as myself.

Wohlstand:
>  Imposing the setting on players would make for an unpleasant experience for those who have above-average sensitivity to sounds.

That looks more like a complication and supporting a sort of "episode author vs its players" war :thinking: That shouldn't be. ~~Added cockroaches. Added anti-cockroach protection. Then, added anti-anti-cockroach protection to let cockroaches keep going. Then improving anti-cockroach protection to stop them go in any way, etc. etc. etc. etc.~~

So, really:
- **ALL** currently existing episodes just don't have this setting being used. This means players will be able to set the thing as they want with no prevention.
- Episode maker OR keeps default behavior that relays on global settings, OR, makes own settings enforced to make sure episode will behave as imagined.  ~~Otherwise, somebody would put a custom "got item" sound with a pre-recorded mix of "player grow" and "got item" which is a much worse case than just a small setting~~. This is a special case that may be used in very rare cases, and the episode maker should be sure of that.

ds-sloth:
Meta-note: we're spending a long time arguing here but I think that is good because the solution we find can be applied to all the other cases (Yoshi coin display, stars on world map, etc...)

0lhi:
> Meta-note: we're spending a long time arguing here

True. At Christmas, too. It is a bit ridiculous.

> I think that is good because the solution we find can be applied to all the other cases

You are right. We should find a solution that will prevent discussions like these in the future. Here's my suggestion:

compat.ini settings should be changeable by the Player from within TheXTech. Either within the Pause Menu, or (better imo) a dedicated Menu for settings stuff.

The changes made by the Player would be saved at:

_/settings/gamesaves/episodename/world-wld(x)-compat1.ini_

In the Pause/Settings Menu, there would be an option to revert the compat settings back to the value set by the Episode Author.

Does that sound good?

Wohlstand:
> /settings/gamesaves/episodename/world-wld(x)-compat1.ini

That would go as the new separated feature which allows players to redefine compat.ini features on the fly, without modification of episode data itself (that allows them to stay fine on read-only FS)

0lhi:
> That would go as the new separated feature which allows players to redefine compat.ini features on the fly

Yeah. It will likely take some time until this feature is added. For now, we should go with my approach, and once the Player can change compat values, ```disable``` can be changed to have the same effect as ```allow```.

Wohlstand:
So, this one I'll just turn into the simple state ([as I explained here](https://github.com/Wohlstand/TheXTech/pull/198#issuecomment-1000911270)), and then, just have the compat-ini override menu which will be just fine. Gonna make fixes...

Wohlstand:
So, compat.ini will have:
- `unspecified` (or `allow`) - default value which will prefer global settings
- `enable` to force enable the feature at episode/level side
- `disable` to force disable the feature at episode/level side

ds-sloth:
@0lhi I want to understand your suggestion better because it sounds *really* cool.

So, player has their own general purpose settings from the main menu. But they also have the specially overridden settings per episode and per level. These use the same format as compat.ini, compat.ini is used for them by default, and when they are present, they override compat.ini. I think that we should also unify the compat.ini format and the settings. When any settings are absent, it follows the previous level of settings.

The final order of dominance is: player per-level settings > creator level settings > player world settings > creator world settings > player global settings > default global settings.

Does that seem appropriate?

0lhi:
> @0lhi I want to understand your suggestion better because it sounds really cool.

^-^

>So, player has their own general purpose settings from the main menu. But they also have the specially overridden settings per episode and per level. These use the same format as compat.ini, compat.ini is used for them by default, and when they are present, they override compat.ini. [...] When any settings are absent, it follows the previous level of settings.

Correct!

> I think that we should also unify the compat.ini format and the settings.

Yes, I can see how that would be beneficial. While at it, we should probably add more seperators to compat.ini, for example one that is dedicated to the [PRE-SMBX64](https://github.com/Wohlstand/TheXTech/issues/185) stuff, and one for TheXTech's bugfixes like #72.

> The final order of dominance is:

With two adjustments:

**Speedrun-Mode 2/3/4(?)** > player per-level settings > creator level settings > player world settings > creator world settings > **target version settings (#163)** > player global settings > default global settings.

Wohlstand:
Okay, I made next:
- settings/thextech.ini has just a simple boolean, no need for any overhead:
```ini
[effects]
enable-player-grow-with-got-item = false ; true/false, false is default
```
- compat.ini has `unspecified` (prefer global by default), `disable` to force-disable at level/episode-side, and `enable` to force-enable at level/episode-side:
```ini
[effects]
enable-player-grow-with-got-item = enable ; unspecified (default), enable, disable
```

No need for extra overhead here, because of the ability for users to non-destructively tune the episode compatibility settings via the menu, can successfully work with the current backend (means, once user-custom compatibility setup are defined, it will be read after episode/level compat.ini will proceed, so, user-side setup will override episode-site). That's for the future when features will be in work, but for now, we don't need overhead here.

0lhi:
Can you change the entry to something like ```sfx-player-grow-with-got-item``` or ```sound-player-grow-with-got-item```?

```enable-player-grow-with-got-item``` is a rather unhelpful description.

**Comment by Wohlstand** [comment]
@0lhi , sure, and done

**Comment by Wohlstand** [comment]
Now format is:

- settings/thextech.ini has just a simple boolean, no need for any overhead:
```ini
[effects]
sfx-player-grow-with-got-item = false ; true/false, false is default
```
- compat.ini has `unspecified` (prefer global by default), `disable` to force-disable at level/episode-side, and `enable` to force-enable at level/episode-side:
```ini
[effects]
sfx-player-grow-with-got-item = enable ; unspecified (default), enable, disable
```

Wohlstand:
Thanks @0lhi for test and confirmation it works fine 🦊👍

ds-sloth:
Small question: what is the `compat.h:68` "bool ignore_got_item_setting"? I thought this was removed.

Wohlstand:
It's now a garbage... forgot! Need to remove it, yeah,  thanks for the point! :-)

PR #199, opened 2021-10-02
[VITA] Fix for Mangled Render Rects, Add Support for Drawing Unfilled Rects.

suicvne:
I present two humble commits based off of the `devel` branch that fix 2 Vita issues on my end.
All files in this patch edit only Vita specific files luckily :+1: 

## First issue
image

As pointed out by ds-sloth, Vita clearly had a problem with its `FrmMain::renderRect` function. Upon closer inspection of my code, I discovered that I was naively normalizing coordinates passed into `renderRect` against the current `viewport_(width/height)`. Thus, resulting in the stretched screen. 

After fix, on real hardware:
VITA_VIEWPORT_FIXED

## Second Issue
image

The second issue fixed in this PR is the Vita's lack of unfilled rectangles. This made me realize some of the short comings of my specialized vgl_renderer.c object, however I found a way to make it work *for now* and not throw off the stride of my glDrawArrays call. 
However, marking draw calls as filled or unfilled was the difficult part. Once I found a decent way to do that, it was a simple matter of setting the draw mode to `GL_LINE_LOOP` (good ref) and swapping the last two vertices.

After fix, again on real hardware:
image

I have tested these fixes on my personal Vita with `-DCMAKE_BUILD_TYPE=Release` and they work as intended. In Discord, I aluded to the right side borders disappearing on single screen levels. I believe this is due to a tainted SDK setup on my MacBook Pro, where I was working when I made that discovery. Re-building the fixes from scratch on my Linux machine seems to produce much better results.

Here is a pre-compiled VPK in `Release` configuration for anyone to try out:
[thextech_Release.zip](https://github.com/Wohlstand/TheXTech/files/7272244/thextech_Release.zip)

### DIFF
--- a/src/vita/frm_main.cpp
+++ b/src/vita/frm_main.cpp
@@ -822,13 +822,13 @@ void FrmMain::renderRect(int x, int y, int w, int h, float red, float green, flo
 {
     (void)filled;
     // TODO: Filled or not?
-    PGE_RectF rect = __NormalizeToGL(x + viewport_offset_x, y + viewport_offset_y, w, h, viewport_w, viewport_h);
+    PGE_RectF rect = __NormalizeToGL(x + viewport_offset_x, y + viewport_offset_y, w, h, ScreenW, ScreenH);
     Vita_DrawRectColor(rect.left, rect.top, rect.right - rect.left, rect.bottom - rect.top, red, green, blue, alpha);
 }
 
 void FrmMain::renderRectBR(int _left, int _top, int _right, int _bottom, float red, float green, float blue, float alpha)
 {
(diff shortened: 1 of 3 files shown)


suicvne:
To re-add Vita to the CI team:

image

The change would be as simple as changing the highlighted line for `devel` or adding `devel` to the list :+1:

ds-sloth (approved):
Looks great. I think that so long as you are specifically modifying files in the Vita subfolder you can directly push them to devel. You are the developer for the Vita port after all. If there is anything you want to commit to devel that modifies the rest of our code, please request review in this way for now, but maybe in the future it will be better for you to directly commit your changes and just notify us when things are changed.

Wohlstand (approved):
For me, looks fine

PR #200, opened 2021-10-06
Small update the texture rendering code

Wohlstand:
- Remove all integer-based texture render calls (they were just unused externally)
- Use floating-point based calls, as SDL2 inside self do call floating-point based calls through integer-based
- Don't call the color changes per texture when that's already matches a set configuration

### DIFF
--- a/src/frm_main.cpp
+++ b/src/frm_main.cpp
@@ -1706,7 +1706,8 @@ void FrmMain::renderCircle(int cx, int cy, int radius, float red, float green, f
     cx += viewport_offset_x;
     cy += viewport_offset_y;
 
-    for(double dy = 1; dy <= radius; dy += 1.0)
+    double dy = 1;
+    do //for(double dy = 1; dy <= radius; dy += 1.0)
     {
         double dx = std::floor(std::sqrt((2.0 * radius * dy) - (dy * dy)));
         SDL_RenderDrawLine(m_gRenderer,
@@ -1714,6 +1715,7 @@ void FrmMain::renderCircle(int cx, int cy, int radius, float red, float green, f
                            int(cy + dy - radius),
                            int(cx + dx),
                            int(cy + dy - radius));
+
         if(dy < radius) // Don't cross lines
         {
             SDL_RenderDrawLine(m_gRenderer,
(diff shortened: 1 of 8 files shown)


ds-sloth:
As a side note: I think we should switch our rendering code from class-based to static-variable based. (`frmMain.renderRect` -> `Render::rect`) This would allow the different clients to implement different utility functions and static variables without declaring them in the header.

ds-sloth:
Because 3DS's tint mode is incompatible with SDL's, 3DS only supports shadow mode but not other tints. That's why I was thinking we should include them as separate support levels.

suicvne:
My comments:

- I like the changes presented. As Wohl stated, Vita renderer is float-based since internally it uses GL ES. So the updates should play nice. By the time the renderer sees any coordinates, they are already normalized from -1.0 -> +1.0 

- I also like ds-sloth’s presented idea of formalizing the render interface, and abstracting it away from the idea of being bundled in frmMain (or at least *named* frmMain). 

Vita should be compatible with the changes mentioned with little fuss, my only real concern is that my vertex positions and shader parameters are plain `float`s instead of `double`s. I could truncate, but precision loss is a concern. Not sure what/if there would be a speed hit would be for the increased vertex/shader parameter positions size. Only one way to find out!

Wohlstand:
> I also like ds-sloth’s presented idea of formalizing the render interface, and abstracting it away from the idea of being bundled in frmMain (or at least named frmMain).

That what was made at Moondust Engine already, and later really I will need to upgrade the whole thing, so, put this into the plan. :wink:

Wohlstand:
So, am I fine to merge this now?
All about changing the class into static interface for graphics, will be done separately, it's major work that better to do after 1.3.5.2

ds-sloth:
Yes, let's merge. For the future static interface work, do note that I am moving almost all events-related code out of `frmMain` and into the new `Controls` interface, so the new `frmMain` can be entirely dedicated to loading and rendering textures.

Wohlstand:
Right, the FrmMain will be splitted-out into role specific units which can be managed among various platforms.

Wohlstand:
@ds-sloth , @suicvne , I made the issue to keep noted about the FrmMain rework at the future

PR #229, opened 2021-12-18
Big clean-up

Wohlstand:
- Got rid of all x == true and  y == false -like conditions were annoying
- Fixed the dozen of warnings and bugs reported by PVS Studio and CLion
- Using more references than direct array keys

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -82,3 +82,7 @@ cmake-build-*/*
 .vscode/*
 .vs/*
 CMakeSettings.json
+
+# PVS-Studio related stuff
+*.PVS-Studio.*
+.PVS-Studio/
--- a/lib/AppPath/app_path.cpp
+++ b/lib/AppPath/app_path.cpp
@@ -1,20 +1,21 @@
 /*
- * Moondust, a free game engine for platform game making
- * Copyright (c) 2014-2021 Vitaly Novichkov <admin@wohlnet.ru>
+ * TheXTech - A platform game engine ported from old source code for VB6
  *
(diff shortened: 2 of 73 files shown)


ds-sloth:
As I said in the Discord, there may be a small bug introduced somewhere in this merge because there are so many changes. It might make it harder to merge changes from the other branches, also.

But, the new code is more readable and better in general. It will make the project more maintainable. I am happy to approve this merge.

ds-sloth (approved):

suicvne (commented):
Wow, there's *a lot* of code changes in this pull. It's a huge one, but from what I'm seeing it's mostly semantics updates. I'm seeing a couple of spots where direct variable passes are replaced by pass-by-reference. Those are the only spots I'm concerned about with this PR for compatibility reasons are the pass-by-reference changes. 

Much of these changes make sense and are clean-ups I would want to perform in my own projects. So I give my approval

ds-sloth:
I believe the Vita CI is not enabled on this branch since it doesn't affect any devel code. Wohlstand shared in Discord that our plan is to release 1.5.3 and then begin porting features from devel to stable, watching for compatibility as each feature is ported.

I agree that we should watch the references. I remember introducing a subtle bug when I tried to change a string pass-by-value to a pass-by-const ref in my own code (probably in the new editor, can't remember right now). It's a good optimization, especially for any frequently-called functions, but let's watch it and we will know where to look if any bugs are introduced.

Wohlstand:
> Vita CI is not enabled on this branch

That is, Vita CI is enabled for `devel` branch only

PR #230, opened 2021-12-19
Use CMake's FindBacktrace instead of #include <execinfo.h>

capezotte:
On some Linux distributions, such as Alpine, the `backtrace` function is not present in the standard C library, causing the build to fail.

This will make it possible to build the engine on these niche distributions, and (untested) allow BSD users to produce backtraces.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,8 +88,9 @@ include(version.cmake)
 include(cmake/git_version.cmake)
 
 
-configure_file(lib/CrashHandler/backtrace.h.in lib/CrashHandler/backtrace.h)
-include_directories(${CMAKE_CURRENT_BINARY_DIR})
+configure_file(lib/CrashHandler/backtrace.h.in
+               generated-include/lib/CrashHandler/backtrace.h)
+include_directories(${CMAKE_CURRENT_BINARY_DIR}/generated-include)
 
 #if(WIN32 OR EMSCRIPTEN OR HAIKU)
 #    set(USE_SYSTEM_LIBS_DEFAULT OFF)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,6 +6,7 @@ include(CheckCXXCompilerFlag)
 include(ExternalProject)
 include(CheckLibraryExists)
 include(GNUInstallDirs)
+include(FindBacktrace)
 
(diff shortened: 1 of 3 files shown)


Wohlstand:
Good job 🦊👍
I'll review and test the thing once I get my home

CMakeLists.txt
```diff
@@ -87,6 +88,9 @@ include(version.cmake)
 include(cmake/git_version.cmake)
 
 
+configure_file(lib/CrashHandler/backtrace.h.in lib/CrashHandler/backtrace.h)
+include_directories(${CMAKE_CURRENT_BINARY_DIR})
```
Wohlstand, line 92: I highly recommend to don't making binary root being included: Please make the sub-directory "generated-include" for such stuff (maybe something also could be done by this way)

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 211cebf..292a52e 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,8 +88,9 @@ include(version.cmake)
 include(cmake/git_version.cmake)
 
 
-configure_file(lib/CrashHandler/backtrace.h.in lib/CrashHandler/backtrace.h)
-include_directories(${CMAKE_CURRENT_BINARY_DIR})
+configure_file(lib/CrashHandler/backtrace.h.in
+               generated-include/lib/CrashHandler/backtrace.h)
+include_directories(${CMAKE_CURRENT_BINARY_DIR}/generated-include)
 
 #if(WIN32 OR EMSCRIPTEN OR HAIKU)
 #    set(USE_SYSTEM_LIBS_DEFAULT OFF)
```

Wohlstand (changes requested):
All looks fine, but I made one minor note and a small hint :wink:

CMakeLists.txt
capezotte: I've just applied the suggested patch to the PR, good catch.

However, I'm not sure I understand what you mean by
> maybe something also could be done by this way

Are there any changes besides the suggested one I'm not seeing?

Wohlstand: I meant generally, when making any generated files, please put them at dedicated directory to guarantee no alien garbage (generated by build) will be messed up at the place.

All fine now, I merged 🦊👍

PR #233, opened 2021-12-27
Screen fading between levels and on world map

Wohlstand:
This work adds the screen fade effect to make smooth transition effect between levels and the rectangular fade effect at world map to represent SMB3 behavior. Also, changed the code of the enter scene (so called "Game Thing" by Redigit): I made sure it will use custom graphics of the level or hub. So, it will be shown after level and all dependent stuff completely loaded. Also, I made it contain built-in timer to make smooth fade appearance and waiting without stuck the whole loop. I think, this stuff is complete, however, next cases were left:
- [x] Attempt to make the fade out after players got kicked off the level by foes
- [x] Attempt to make simplest transition while going between warps


Wohlstand:
Okay, the most of works here is done if don't count the "nobody alive" case.

To test per-warp fade, you should manually add the "TE:<x>;" field at warp in the LVLX file as Moondust Editor doesn't support this field yet:
[Uploading Transition test.lvlx.zip…]()
```
Values are next:
- 0 - None
- 1 - Scroll (not implemented yet, will fall back into None)
- 2 - Fade
- 3 - Circle fade
- 4 - Flip H
- 5 - Flip V
```

src/globals.h
```diff
@@ -906,9 +907,11 @@ struct Warp_t
     bool noPrintStars = false;
     bool noEntranceScene = false;
     bool cannonExit = false;
+    bool stoodRequired = false; // Require player stood on the ground to enter this warp
     double cannonExitSpeed = 10.0;
     std::string eventEnter;
     std::string StarsMsg;
+    int transitEffect = 0;
```
ds-sloth, line 914: Use an enum instead? :)

src/graphics/gfx_enter_screen.cpp
```diff
@@ -18,56 +18,133 @@
...
 #include "../globals.h"
 #include "../graphics.h"
 #include "../player.h"
+#include "../frame_timer.h"
+#include "../screen_fader.h"
+#include "pge_delay.h"
 
 
-void GameThing()
+static void initPlayers(Player_t tempPlayer[2])
```
ds-sloth, line 31: Can you instead use `maxLocalPlayers` here? Not sure if C arrays support constants here, but the abstract-controls code is trying to allow for >2 players in the future.

src/graphics/gfx_enter_screen.cpp
```diff
@@ -18,56 +18,133 @@
...
 {
-    int A = 0;
-//    int B = 0;
-    Player_t tempPlayer[3];
+    for(int A = 0; A < numPlayers; ++A)
```
ds-sloth, line 33: Dangerous, what if player used `supermario128` and numPlayers > maxLocalPlayers?

src/graphics/gfx_update.cpp
```diff
@@ -862,6 +863,7 @@ void UpdateGraphics(bool skipRepaint)
 //                End With
 //            Next A
 //        Else
+        if(numBackground > 0)
```
ds-sloth, line 866: I remember there being an error here in my branch (either `devel` or `wip-abstract-controls`) too, but I think I fixed it in a different way. We can see when we merge the branches.

src/main/game_globals.h
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+#ifndef GAME_GLOBALS_H
+#define GAME_GLOBALS_H
+
+#include "../screen_fader.h"
+#include "../range_arr.hpp"
+
+extern ScreenFader g_levelScreenFader;
+extern RangeArr<ScreenFader, 0, 2> g_levelVScreenFader;
```
ds-sloth, line 28: Why max of 2 here?

src/main/game_loop.cpp
```diff
@@ -88,19 +107,19 @@ void GameLoop()
         UpdateEffects();
         speedRun_tick();
         UpdateGraphics();
+        updateScreenFaders();
```
ds-sloth, line 110: Can you put these calls in `UpdateGraphics` instead? Maybe more maintainable to have fewer special calls in the main loops

ds-sloth (approved):
These look good. I had a few minor suggestions but I am happy with this. I may bypass some of these new options on the low-resource clients where drawing the circle window is not easy, but it is great that we have them for our high-resource clients.

I had two comments about supporting `n` players instead of <=2 players, and one comment about moving the fader calls into `UpdateGraphics`.

I also like the cleanup you did for `npc_hit.cpp`, didn't see a way to write a full-file comment.

I do want to check that it is possible to disable the new fade effects. Is that right?

src/globals.h
Wohlstand: It's already enum just an integer container here (to don't include PGE-FL structures here at "globals.h"), in the code you'll find direct enum usage :wink:

src/graphics/gfx_enter_screen.cpp
Wohlstand: Good note!

src/graphics/gfx_enter_screen.cpp
Wohlstand: Lemme try...

Wohlstand: But, you are absolutely right, the tempPlayer[] will result out of range

Wohlstand: So, this condition should be right:
```cpp
A < numPlayers && A < maxLocalPlayers
```

src/main/game_globals.h
Wohlstand: Because of vScreen-related arrays, they are all has from 0 to 2 range

src/main/game_loop.cpp
Wohlstand: This thing is a backend part, and I think it's better to separate render and other workflow-related updates

src/graphics/gfx_enter_screen.cpp
Wohlstand: Fixed :+1:

src/globals.h
ds-sloth: Ah right, I usually use C++-style enum classes so I was expecting to see that. This is good too! :)

src/main/game_loop.cpp
ds-sloth: I think the screen faders are similar to the screen shake, which does live in `UpdateControls`, but if you prefer it here I am good with that also.

Wohlstand: I think we could try something later to tweak this and clean up the whole stuff: in the future, I plan to reorganize the whole rendering making it use the Z-value concept, which should allow the flexible rendering of objects like at Moondust and at X2 is possible. This thing is required for a multi-layered backgrounds system which already requires some Z-value support to allow drawing layers in many ways:
- under everything like usually now
- among in-scene objects
- over everything (making sort of decorations like plants, etc.)

Therefore, this thing could be improved later :wink:

ds-sloth: This sounds exciting, and you can see in the devel branch one approach to separate the logic components of `UpdateGraphics` from the actual rendering. The NPC draw queue class is basically a simple (and high performance, since no sorting is required) Z-value system. If we use a sorting-based system instead of a multi-queue system we should be careful to ensure high performance on the systems like Vita and 3DS that still need optimization.

src/main/game_globals.h
```diff
@@ -0,0 +1,29 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef GAME_GLOBALS_H
+#define GAME_GLOBALS_H
+
+#include "../screen_fader.h"
+
+extern ScreenFader g_levelScreenFader;
```
suicvne, line 27: Why here and not in `globals.cpp`?

src/screen_fader.h
```diff
@@ -0,0 +1,117 @@
...
+#define SCREEN_FADER_H
+
+#include "frm_main.h"
+#include "fader.h"
+#include "global_constants.h"
+
+extern FrmMain frmMain;
+
+
+struct ScreenFader
```
suicvne, line 27: Clean 👌🏼 a good example of a simple add-on component to the base engine

src/player/player_update.cpp
```diff
@@ -2218,28 +2224,43 @@ void UpdatePlayer()
...
                         LevelBeatCode = 3;
                         EndLevel = true;
                         LevelMacro = LEVELMACRO_OFF;
                         LevelMacroCounter = 0;
-                        frmMain.setTargetTexture();
-                        frmMain.clearBuffer();
-                        frmMain.repaint();
+                        g_levelScreenFader.setupFader(4, 0, 65, ScreenFader::S_FADE);
+
+                        while(!g_levelScreenFader.m_full && GameIsActive)
```
suicvne, line 2247: Maybe blocks like this could be placed in a separate "WaitForFader" method? I saw this pattern repeated a lot and even if the same one is not used multiple times in code, breaking this out into its own method (even `static inline`'d) could help readability some.

src/main/game_globals.h
Wohlstand: Because I desired to start the gradual organizing of the thing and making "globals" thing stop to grow, probably later going to split it down into multiple parts (mainly, public structures such as player, block, NPC, etc., etc.)

src/player/player_update.cpp
Wohlstand: Good point

Wohlstand: Really, could be better, also, you just now gave me an idea to solve the "everyone dead" task! So, I made this call and I re-used it at "EveryonesDead()" call

Wohlstand: Done, should be better now :wink: 
Also, fade now works when all players have gone

Wohlstand:
> I do want to check that it is possible to disable the new fade effects. Is that right?

Whoops, missed up this, didn't implement that ability yet, I would make it later as a config option

src/main/game_loop.cpp
Wohlstand: So, that's a topic for future discussing :wink:

src/screen_fader.h
Wohlstand: Thanks :fox_face:

Wohlstand:
> I may bypass some of these new options on the low-resource clients were drawing the circle window is not easy

There could work a simple workaround: 
- make the square texture which has the closed size to screen
- fill it with the circle window
- while drawing a thing, draw the window, and 4 rectangles around

By the fact, Super Princess Peach for NDS already uses the circle window effect easily :eyes:
What about New Super Mario Bros. for the same NDS?

Wohlstand:
Okay, gonna merge this now.
- Everything would be needed we can apply on the top quickly
- I gonna fix up some other cases were are needs both things merged

Wohlstand:
I made the discussion against the circle window:
<https://github.com/Wohlstand/TheXTech/discussions/235>

PR #234, opened 2021-12-29
Merge gameplay recordings

ds-sloth:
The results here are not perfect (because not all divergences from vanilla game are found, TheXTech is not perfectly compatible with SMBX yet in spite of our efforts) but I do not think there are any errors introduced here and I do fix one or two divergences from the vanilla game.

Please consult my work; the major change is the new random engine. And, of course, the fact that we are now able to track the gameplay recording for any sorts of changes and divergence. Even though the SMBX-TheXTech compatibility is not perfect, we can also test TheXTech-TheXTech compatibility as we implement new features. If nothing here seems displeasing to you, I would advise merging it immediately so that we can perform bisection testing on bugs in the future.

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -1,0 +1,10 @@
+#ifndef CONFIG_H
+#define CONFIG_H
+
+extern struct Config_t
+{
+    //! Record gameplay data
+    bool    RecordGameplayData = false;
+} g_config; // main_config.cpp
+
+#endif // CONFIG_H
--- a/src/effect.cpp
+++ b/src/effect.cpp
@@ -574,7 +574,7 @@ void UpdateEffects()
                     if(e.Frame >= 4)
                         e.Frame = 0;
                 }
-            if(e.Type == 148 && (dRand() * 10.0) > 8.0)
+            if(e.Type == 148 && iRand(10) >= 8)
             {
                 NewEffect(77, e.Location, 3);
                 Effect[numEffects].Location.SpeedX = dRand() * 3 - 1.5;
@@ -1305,7 +1305,7 @@ void NewEffect(int A, const Location_t &Location, float Direction, int NewNpc, b
                 ne.Location.X += ne.Location.SpeedX * 3;
                 ne.Location.Y += ne.Location.SpeedY * 3;
 
-                ne.Frame = iRand() % 4;
+                ne.Frame = iRand(4);
                 ne.Type = A;
             }
         }
@@ -1547,9 +1547,9 @@ void NewEffect(int A, const Location_t &Location, float Direction, int NewNpc, b
         ne.Location.SpeedY = -2 - dRand() * 10;
         ne.Location.SpeedX = dRand() * 8 - 4;
         ne.Frame = 0;
-        if((iRand() % 2) == 1)
+        if(iRand(2) == 0)
             ne.Frame = 7;
-        ne.Frame += iRand() % 7;
+        ne.Frame += iRand(7);
         ne.Life = 300;
         ne.Type = A;
     }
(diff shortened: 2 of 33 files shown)


src/config.h
```diff
@@ -0,0 +1,10 @@
+#ifndef CONFIG_H
+#define CONFIG_H
+
+extern struct Config_t
+{
+    //! Record gameplay data
+    bool    RecordGameplayData = false;
+} g_config; // main_config.cpp
```
Wohlstand, line 8: A good structure, I will need to move here all other global settings were toggled by settings/thextech.ini, should be more organized! :fox_face: :+1:

ds-sloth: Glad you like this! :)

I am also thinking in the long term we may be able to combine the compat and config data but that is not something to do now.

src/effect.cpp
```diff
@@ -1547,9 +1547,9 @@ void NewEffect(int A, const Location_t &Location, float Direction, int NewNpc, b
         ne.Location.SpeedY = -2 - dRand() * 10;
         ne.Location.SpeedX = dRand() * 8 - 4;
         ne.Frame = 0;
-        if((iRand() % 2) == 1)
+        if(iRand(2) == 0)
```
Wohlstand, line 1550: Why 0 and not 1? But seems, doesn't matter, just the 1/2 chance this will be true.

src/game_main.cpp
```diff
@@ -229,7 +230,14 @@ int GameMain(const CmdLineSetup_t &setup)
...
         LevelSelect = false;
-        FullFileName = setup.testLevel;
+        if(Files::hasSuffix(setup.testLevel, ".rec"))
+        {
+            Record::LoadReplay(setup.testLevel);
+        }
+        else
+        {
+            FullFileName = setup.testLevel;
+        }
```
Wohlstand, line 240: My little tip: Feel free to use braceless condition if not so complex code inside:
```cpp
        if(Files::hasSuffix(setup.testLevel, ".rec"))
            Record::LoadReplay(setup.testLevel);
        else
            FullFileName = setup.testLevel;
```

Also, I highly recommend you to put spaces between condition begin and previous code, also after last brace of condition to make it look more readable:
```cpp
code();

if(cond)
    ko1();
else
    ko2();

code2();
```

Little example that makes harder for quick view and may cause possible misses without proper attention:
```cpp
if(cond)
    ko1();
if(cond)
    ko2();
else if(cond)
    ko2();
if(cond)
    ko2();
else if(cond)
    ko2();
else if(cond)
    ko2();
```

So, better to write as:
```cpp
if(cond)
    ko1();

if(cond)
    ko2();
else if(cond)
    ko2();

if(cond)
    ko2();
else if(cond)
    ko2();
else if(cond)
    ko2();
```

Wohlstand: P.S. There is a lot of code after Redigit's VB6 which is confusing because of that

src/effect.cpp
ds-sloth: Just silly micro-optimization; I think that 1 translates to a subtract and a test, and 0 translates to just a test. Does that sound right to you?

src/game_main.cpp
ds-sloth: That is great advice, thank you Wohlstand!

src/effect.cpp
Wohlstand: ye

Wohlstand:
BTW, Ouch, didn't pay attention, but the RNG code caused MSVC to build breakage (which is used to build Windows ARM64 builds), so, gonna fix up this...

PR #241, opened 2022-01-03
Split FrmMain into abstract sub-systems 

### LINKED ISSUE #210: Split the FrmMain class into many role-specific things
From a discussion for pull-request #200, we need to improve the platforms support between different sub-systems managed. Current FrmMain is a god thing that unites the graphics and controls at one big pagan god of the Game.

So, instead of FrmMain class, make:
- The static interface for graphical rendering (similar to that one made at Moondust Engine to support OpenGL of different versions and the SDL Render as a fallback).
- The similar static interface(s) to manage controls (joysticks, events, keyboard, and mouse).

Right now for the 1.3.5.2 it's too radical change, so, let's put it for future work.

Wohlstand:
Closes #210


Wohlstand:
@ds-sloth, please check the ability to merge this with abstract-controls too

suicvne:
This looks excellent and easy to work with and understand 👌🏼

ds-sloth (commented):
Most of these changes look fantastic. I have one question and one suggestion.

Question: I see you making a lot of XRender calls, but I can't find the file where they are defined. Could you direct me to it?
Suggestion: define StdPicture in the XRender namespace, so that platform-specific renderers can use their own specific versions of StdPicture there. Let it be an incomplete declaration everywhere else. (Or, if other places need to access width or height, let those be defined in an abstract base class.) If StdPicture were defined the way it is now, I would wrap it in an `#ifndef __3DS__` guard and redefine it because it contains too many RAM intensive C++ types at present.

Wohlstand:
> I see you making a lot of XRender calls

You need the `"core/render.h"` for them

Wohlstand:
> StdPicture

I made it is the same for all platforms, just it may have different fields specific for backends (like an integer for OpenGL texture ID, or pointer for SDL_Texture, or something also), and I don't want to make it more complicated and contain duplicated declarations of it... 
By the way, at Moondust Engine, the `PGE_Texture` is the same for all graphical backends, and the SDL-based backend tracks every texture instance in its internal array, and it gives a new id of that array while loading some. When removing something, it doesn't make an offset of the array, and next time, just re-use free IDs when loading any new texture again.

ds-sloth:
Just be aware that StdPicture will end up looking like this unless we give it different definitions per backend: https://github.com/Wohlstand/TheXTech/blob/devel/src/std_picture.h

Wohlstand:
> Just be aware that StdPicture will end up looking like this unless we give it different definitions per backend: https://github.com/Wohlstand/TheXTech/blob/devel/src/std_picture.h

I may guess, the sort of sub-structure can be added then:
```cpp
struct StdPictureData{ ... };

struct StdPicture
{
     ...
     StdPictureData data;
}
```
So, then put that stuff at the another header

So:
- all shared data kept at the same StdPicture
- all platform-specific data will stay on sub-structure

Wohlstand:
btw, looks like I should make the separated sub-system for image loading, mainly:
- Everything uses the same image loading mechanism, same image format libraries, etc.
- Very minor set of specific platforms such as 3DS, needs a majorly different loader
So, then, StdPicture will also have the `StdPictureLoader` sub-structure related to the loading process only, and the different backend part which proceeds the image loading

PR #243, opened 2022-01-06
Updated cheats sub-system

### LINKED ISSUE #61: Allow using cheats by software keyboard on mobile devices
Some levels (such as "Superb Demo 128" at "Prelude To The Stupid") do require users to type cheat code to be passed. Right now on mobile devices that is impossible without the USB-OTG plugged hardware keyboard. There is a possible solution to add the calling for the software keyboard where users will be able to type any cheats they want. That should allow them to easily pass levels that require the user to cheat, or simply allows the user to use chats while casual playthrough.

### LINKED ISSUE #62: Implement the cheat codes renaming by gameinfo.ini
The "Prelude To the Stupid" is a big example where the game had used the modified executable file where all cheat codes got renamed into different names. Some levels (such as "Superb Demo 128") do require the user to type the "superbdemo128" (the replacement for "supermario128") and then, "raomode" (1player) to turn the game back into the normal 1-player mode. The fact cheat codes were mentioned in the game in their renamed form, may confuse users who do use the engine that uses original cheats being here.

Wohlstand:
- All spaghetti of cheat codes was exploded into separated static functions
- The list of all known cheats is now dynamic and can be extended or modified
- gameinfo.ini now supports adding aliases or renaming for existing cheats (mainly needed for "ASMBXT: PTTS" which expects all cheats renamed in their own way)
- on Android, added the ability to type cheats using Start+KeyboatdShowHide touchscreen button combinations (the special dialog will pop up where you can type the cheat using software keyboard)

P.S. With this update of the cheats system, will be possible to provide the sort of cheats menu, without manual copypasting of cheat codes.

Closes #62

Now, also Closes #61


src/graphics/gfx_enter_screen.cpp
```diff
@@ -143,7 +143,7 @@ static void initPlayers(Player_t tempPlayer[maxLocalPlayers])
 
 static void drawEnterScreen(Player_t tempPlayer[maxLocalPlayers])
 {
-    for(int A = 0; A < numPlayers; ++A)
+    for(int A = 0; A < maxLocalPlayers; ++A)
```
ds-sloth, line 146: Maybe better to do this one with the A < max && A < num pattern? Uncertain how the inner code works though.

Wohlstand: You right

src/main/game_info.cpp
```diff
@@ -23,12 +23,30 @@
...
+    dst.clear();
+    conf.beginGroup(group);
+    {
+        auto ak = conf.allKeys();
+        for(const auto &k : ak)
+        {
+            GameInfo::CheatAlias ca;
+            ca.first = k;
+            conf.read(k.c_str(), ca.second, std::string());
+            dst.push_back(ca);
```
ds-sloth, line 44: I'm worried that the alias could be an empty string. I assume that elsewhere you check for empty aliases? Or is the default is necessarily unused since k was in the keys array?

Wohlstand: Ah, damn, I should just skip adding this if becomes empty...

src/main/cheat_code.cpp
```diff
@@ -1010,739 +1088,1028 @@ void CheatCode(char NewKey)
...
+
+    CheatString.push_back(NewKey);
+    if(CheatString.size() > 23)
+        CheatString.erase(0, 1);
+    oldString = CheatString;
+
+
+    for(const auto &c : s_cheatsListGlobal)
+    {
+        if(CheatString.find(c.key) != std::string::npos)
```
ds-sloth: Can you do a direct suffix check here instead of the find call?

ds-sloth (commented):
Looks good to me. Left comments above. I think this is cleaner and more future proof.

If you keep the old indentation it might be easier for us to merge other branches on top of this but that's no big issue.

Do you know how much memory the cheat array consumes? If more than ~32KB we may want to consider replacing the cheat functions with stubs on very low memory devices.

src/main/cheat_code.cpp
Wohlstand: Good idea

Wohlstand: And, for the best performance, reasonable to use `memcmp()` here

Wohlstand:
> Do you know how much memory the cheat array consumes? If more than ~32KB we may want to consider replacing the cheat functions with stubs on very low memory devices.

The Max length of the cheat string is 25 symbols, I made sure the buffer has 25 characters maximum. The whole thing would take ~2000 bytes

src/graphics/gfx_enter_screen.cpp
Wohlstand: Done, now the check rely on both num players and max players

src/main/game_info.cpp
Wohlstand: Fixed

PR #267, opened 2022-01-31
Low-level backports from devel

ds-sloth:
In this PR I share 3 backported features from the devel branch:

(1) Shared allocation for tree query results -- saves memory and simplifies code, making it easier to add new tree queries without allocating vectors in every function that queries a tree. This will be used in the Block quadtree (or table, whichever we set up).

(2) Objects now reference their layers, events, and strings via indices. In turn, layers have sets containing their members' indices. This dramatically improves performance in levels with multiple moving layers. It will also be used in the Block quadtree (or table, as before). Programmer's note: in general, we will want to use `FindLayer`, `FindEvent`, and `SetS` to convert strings to indices, and `Layer[index].Name`, `Event[index].Name`, and `GetS` to convert indices to strings.

(3) The more efficient `DirListCI` has been ported. Note that it has more calls for specific cases (including existence checks and requesting absolute paths). The many copies of `s_dirEpisode` and `s_dirCustom` have been replaced with `g_dirEpisode` and `g_dirCustom`, declared in `global_dirs.h`.

(Small addition: it was hard to merge 1 and 2 without also updating some of the editor code. I will continue to update this later, but please confirm that my changes do not break Interprocess.)

Most of the changes to autocode are simple and easy, but please take a look at them. I had to remove the hooks to set the `AttLayer`, `Text`, `Layer`, `TriggerActivate`, `TriggerLast`, and `TriggerDead` attributes of `NPC_t` because the referred-to-memory is now a `uint8_t` or `uint16_t` and not a `std::string` as before. As far as I can see, the previous version did not support string assignment either, so I am not worried about this. I did add a `syncLayers_NPC` hook in the appropriate place for when we do reimplement setting the NPC's layer correctly.

These are all well-tested features in devel, but whenever we backport features there is always the risk of mistakes. Please do skim the code and let me know if there are any errors. If not, it should be a pretty simple merge. Thanks!

### DIFF
--- a/lib/Utils/dir_list_ci.cpp
+++ b/lib/Utils/dir_list_ci.cpp
@@ -4,6 +4,10 @@
 #include "strings.h"
 #include "dir_list_ci.h"
 
+#include <cstring>
+#include <algorithm>
+#include <utility>
+
 
 DirListCI::DirListCI(std::string curDir) noexcept
     : m_curDir(std::move(curDir))
@@ -13,8 +17,16 @@ DirListCI::DirListCI(std::string curDir) noexcept
 
 void DirListCI::setCurDir(const std::string &path)
 {
-    m_curDir = path;
-    rescan();
+    if(path != m_curDir)
+    {
+        m_curDir = path;
+        rescan();
+    }
+}
+
+const std::string& DirListCI::getCurDir()
+{
+    return m_curDir;
 }
 
 static void replaceSlashes(std::string &str, const std::string &from)
@@ -38,13 +50,49 @@ static void replaceSlashes(std::string &str, const std::string &from)
     }
 }
 
-std::string DirListCI::resolveFileCase(const std::string &in_name)
+bool DirListCI::existsCI(const std::string &in_name)
 {
-#if 0 //def _WIN32
+    if(in_name.empty())
+        return false;
+
     std::string name;
     replaceSlashes(name, in_name);
-    return name;
-#else
+
+    // For sub-directory path, look deeply
+    auto subDir = name.find('/');
+    if(subDir != std::string::npos)
+    {
+        // THIS BEHAVIOR IS EXPENSIVE AND WASTEFUL, AVOID INVOKING AT ALL COSTS!!!
+        auto sdName = resolveDirCase(name.substr(0, subDir));
+        DirListCI sd(m_curDir + "/" + sdName);
+        return sd.existsCI(name.substr(subDir + 1));
+    }
+
+    // keep MixerX path arguments untouched
+    auto pathArgs = name.find('|');
+    if(pathArgs != std::string::npos)
+    {
+        auto n = name.substr(0, pathArgs);
+        std::string uppercase_string;
+        uppercase_string.resize(n.length());
+        std::transform(n.begin(), n.end(), uppercase_string.begin(),
+            [](unsigned char c){ return std::toupper(c); });
+        auto found = m_fileMap.find(uppercase_string);
+        return found != m_fileMap.end();
+    }
+    else
+    {
+        std::string uppercase_string;
+        uppercase_string.resize(name.length());
+        std::transform(name.begin(), name.end(), uppercase_string.begin(),
+            [](unsigned char c){ return std::toupper(c); });
+        auto found = m_fileMap.find(uppercase_string);
+        return found != m_fileMap.end();
+    }
+}
+
+std::string DirListCI::resolveFileCaseExists(const std::string &in_name)
+{
     if(in_name.empty())
         return in_name;
 
@@ -55,9 +103,14 @@ std::string DirListCI::resolveFileCase(const std::string &in_name)
     auto subDir = name.find('/');
     if(subDir != std::string::npos)
     {
+        // THIS BEHAVIOR IS EXPENSIVE AND WASTEFUL, AVOID INVOKING AT ALL COSTS!!!
         auto sdName = resolveDirCase(name.substr(0, subDir));
         DirListCI sd(m_curDir + "/" + sdName);
-        return sdName + "/" + sd.resolveFileCase(name.substr(subDir + 1));
+        std::string found = sd.resolveFileCase(name.substr(subDir + 1));
+        if(found.empty())
+            return "";
+        else
+            return sdName + "/" + found;
     }
 
     // keep MixerX path arguments untouched
@@ -65,49 +118,129 @@ std::string DirListCI::resolveFileCase(const std::string &in_name)
     if(pathArgs != std::string::npos)
     {
         auto n = name.substr(0, pathArgs);
-        for(std::string &c : m_fileList)
-        {
-            if(SDL_strcasecmp(c.c_str(), n.c_str()) == 0)
-                return c + name.substr(pathArgs);
-        }
+        std::string uppercase_string;
+        uppercase_string.resize(n.length());
+        std::transform(n.begin(), n.end(), uppercase_string.begin(),
+            [](unsigned char c){ return std::toupper(c); });
+        auto found = m_fileMap.find(uppercase_string);
+        if(found != m_fileMap.end())
+            return found->second + name.substr(pathArgs);
     }
     else
-    for(std::string &c : m_fileList)
     {
-        if(SDL_strcasecmp(c.c_str(), name.c_str()) == 0)
-            return c;
+        std::string uppercase_string;
+        uppercase_string.resize(name.length());
+        std::transform(name.begin(), name.end(), uppercase_string.begin(),
+            [](unsigned char c){ return std::toupper(c); });
+        auto found = m_fileMap.find(uppercase_string);
+        if(found != m_fileMap.end())
+            return found->second;
     }
 
-    return name;
+    return "";
+}
+
+std::string DirListCI::resolveFileCase(const std::string &in_name)
+{
+#if 0
+    return in_name; // no need on Windows
+#else
+    if(in_name.empty())
+        return in_name;
+
+    std::string found = resolveFileCaseExists(in_name);
+
+    // if not found, overwrite with the replace-slash version of the name
+    // and return
+    if(found.empty())
+        replaceSlashes(found, in_name);
+
+    return found;
 #endif
+}
+
+std::string DirListCI::resolveFileCaseAbs(const std::string &in_name)
+{
+#if 0
+    return m_curDir + "/" + in_name; // no need on Windows
+#else
+    if(in_name.empty())
+        return in_name;
+
+    std::string found = resolveFileCaseExists(in_name);
+
+    // if not found, overwrite with the replace-slash version of the name
+    // and return
+    if(found.empty())
+        replaceSlashes(found, in_name);
+
+    return m_curDir + "/" + found;
+#endif
+}
+
+std::string DirListCI::resolveFileCaseExistsAbs(const std::string &in_name)
+{
+    if(in_name.empty())
+        return in_name;
+
+    std::string found = resolveFileCaseExists(in_name);
+
+    // if not found, overwrite with the replace-slash version of the name
+    // and return
+    if(found.empty())
+        return found;
+
+    return m_curDir + "/" + found;
 }
 
 std::string DirListCI::resolveDirCase(const std::string &name)
 {
-#if 0 //def _WIN32
+#if 0
     return name; // no need on Windows
 #else
     if(name.empty())
         return name;
 
-    for(std::string &c : m_dirList)
-    {
-        if(SDL_strcasecmp(c.c_str(), name.c_str()) == 0)
-            return c;
-    }
+    std::string uppercase_string;
+    uppercase_string.resize(name.length());
+    std::transform(name.begin(), name.end(), uppercase_string.begin(),
+        [](unsigned char c){ return std::toupper(c); });
 
-    return name;
+    auto found = m_dirMap.find(uppercase_string);
+    if(found == m_dirMap.end())
+        return name;
+
+    return found->second;
 #endif
 }
 
 void DirListCI::rescan()
 {
-    m_fileList.clear();
-    m_dirList.clear();
+    m_fileMap.clear();
+    m_dirMap.clear();
     if(m_curDir.empty())
         return;
 
     DirMan d(m_curDir);
-    d.getListOfFiles(m_fileList);
-    d.getListOfFolders(m_dirList);
+    std::vector<std::string> fileList;
+    std::vector<std::string> dirList;
+    d.getListOfFiles(fileList);
+    d.getListOfFolders(dirList);
+
+    std::string uppercase_string;
+
+    for(std::string& file : fileList)
+    {
+        uppercase_string.resize(file.length());
+        std::transform(file.begin(), file.end(), uppercase_string.begin(),
+            [](unsigned char c){ return std::toupper(c); });
+        m_fileMap.insert(std::make_pair(uppercase_string, file));
+    }
+    for(std::string& dir : dirList)
+    {
+        uppercase_string.resize(dir.length());
+        std::transform(dir.begin(), dir.end(), uppercase_string.begin(),
+            [](unsigned char c){ return std::toupper(c); });
+        m_dirMap.insert(std::make_pair(uppercase_string, dir));
+    }
 }
--- a/lib/Utils/dir_list_ci.cpp
+++ b/lib/Utils/dir_list_ci.cpp
@@ -106,7 +106,7 @@ std::string DirListCI::resolveFileCaseExists(const std::string &in_name)
         // THIS BEHAVIOR IS EXPENSIVE AND WASTEFUL, AVOID INVOKING AT ALL COSTS!!!
         auto sdName = resolveDirCase(name.substr(0, subDir));
         DirListCI sd(m_curDir + "/" + sdName);
-        std::string found = sd.resolveFileCase(name.substr(subDir + 1));
+        std::string found = sd.resolveFileCaseExists(name.substr(subDir + 1));
         if(found.empty())
             return "";
         else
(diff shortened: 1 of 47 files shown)


src/script/luna/mememu.cpp
```diff
@@ -1305,18 +1308,18 @@ class NPCMemory final : public NpcParent
...
-        insert(0x0000002c, &NPC_t::TriggerActivate);
-        insert(0x00000030, &NPC_t::TriggerDeath);
-        insert(0x00000034, &NPC_t::TriggerTalk);
-        insert(0x00000038, &NPC_t::TriggerLast);
-        insert(0x0000003c, &NPC_t::Layer);
+        // insert(0x0000002c, &NPC_t::TriggerActivate);
+        // insert(0x00000030, &NPC_t::TriggerDeath);
+        // insert(0x00000034, &NPC_t::TriggerTalk);
+        // insert(0x00000038, &NPC_t::TriggerLast);
+        // insert(0x0000003c, &NPC_t::Layer);
```
Wohlstand, line 1315: By fact, these string fields were never modified by autocode, and probably not going to be modded as Autocode was never able to memhack strings. The strings hacking was added at LunaLua later. So, right. For the worst case, you could use the lambda fields instead to add/remove/get string entry later.

src/script/luna/mememu.cpp
```diff
@@ -1407,6 +1411,13 @@ class NPCMemory final : public NpcParent
         }
 
         NpcParent::setValue(obj, address, value, ftype);
+
+        if(address == 0x3C) // Layer, but setting it is not yet supported.
+        {
+            int index = obj - &NPC[0];
+            if(index >= -128 && index <= maxNPCs)
+                syncLayers_NPC(index);
+        }
```
Wohlstand, line 1420: You may just comment out this code with `#if 0` `#endif` block easily until something would appear at any moment is really needed.

lib/Utils/dir_list_ci.cpp
```diff
@@ -55,59 +103,144 @@ std::string DirListCI::resolveFileCase(const std::string &in_name)
...
     {
+        // THIS BEHAVIOR IS EXPENSIVE AND WASTEFUL, AVOID INVOKING AT ALL COSTS!!!
         auto sdName = resolveDirCase(name.substr(0, subDir));
         DirListCI sd(m_curDir + "/" + sdName);
-        return sdName + "/" + sd.resolveFileCase(name.substr(subDir + 1));
+        std::string found = sd.resolveFileCase(name.substr(subDir + 1));
+        if(found.empty())
+            return "";
+        else
+            return sdName + "/" + found;
```
Wohlstand, line 113: This thing is mainly used by music paths as they usually stored at any random folder, and going be relative to the episode directory always.

lib/Utils/dir_list_ci.cpp
```diff
@@ -55,59 +103,144 @@ std::string DirListCI::resolveFileCase(const std::string &in_name)
...
+    std::string uppercase_string;
+
+    for(std::string& file : fileList)
+    {
+        uppercase_string.resize(file.length());
+        std::transform(file.begin(), file.end(), uppercase_string.begin(),
+            [](unsigned char c){ return std::toupper(c); });
+        m_fileMap.insert(std::make_pair(uppercase_string, file));
+    }
+    for(std::string& dir : dirList)
```
Wohlstand, line 239: I'd added the empty line here, otherwise, this looks clobbish

lib/Utils/dir_list_ci.cpp
```diff
@@ -55,59 +103,144 @@ std::string DirListCI::resolveFileCase(const std::string &in_name)
...
     {
-        if(SDL_strcasecmp(c.c_str(), name.c_str()) == 0)
-            return c;
+        std::string uppercase_string;
+        uppercase_string.resize(name.length());
+        std::transform(name.begin(), name.end(), uppercase_string.begin(),
+            [](unsigned char c){ return std::toupper(c); });
+        auto found = m_fileMap.find(uppercase_string);
+        if(found != m_fileMap.end())
+            return found->second;
```
Wohlstand, line 137: I guess, we can simplify this whole block and run only one block for both cases:

- keeping source being constant
- if path args found, set end of length as "begin() + argsBegin", and when not found, just end()

lib/Utils/dir_list_ci.cpp
```diff
@@ -55,59 +103,144 @@ std::string DirListCI::resolveFileCase(const std::string &in_name)
...
 
-    return name;
+    return "";
+}
+
+std::string DirListCI::resolveFileCase(const std::string &in_name)
+{
+#if 0
+    return in_name; // no need on Windows
+#else
```
Wohlstand, line 147: It's time to remove these macros ad this thing is needed for Windows too, mainly because of filenames caching which does work out of FS.

src/globals.h
```diff
@@ -101,6 +101,44 @@ extern int vb6Round(double x);
  */
 extern double vb6Round(double x, int decimals);
 
+
+// utilities for stringindex_t
+extern std::vector<std::string> g_LevelString;
+extern size_t g_numWorldString;
+
+inline std::string GetS(stringindex_t index)
```
Wohlstand, line 109: Maybe return the const referrence?

src/globals.h
```diff
@@ -101,6 +101,44 @@ extern int vb6Round(double x);
...
+{
+    if(index == STRINGINDEX_NONE)
+    {
+        return "";
```
Wohlstand, line 113:
```cpp
return std::string();
```

lib/Utils/dir_list_ci.cpp
ds-sloth: Thank you for the info. I knew that most cases would not invoke the subdir search, but I didn't know when it would be used.

src/script/luna/mememu.cpp
ds-sloth: I will, thanks.

src/script/luna/mememu.cpp
ds-sloth: Cool! So, when (if) we implement LunaLua, that's the time we will need to add these lambda fields.

lib/Utils/dir_list_ci.cpp
Wohlstand: There are many episodes that links music at subdirectories, the significant examples are A2XT games that directly require contributors to put music into sub-directory

lib/Utils/dir_list_ci.cpp
ds-sloth: Yes, that is a great suggestion!

lib/Utils/dir_list_ci.cpp
ds-sloth: Yes, I will do this.

lib/Utils/dir_list_ci.cpp
ds-sloth: Done.

src/globals.h
ds-sloth: I agree that a const ref is usually a good idea. There are two reasons I wanted to use a string instead of a const ref:
- There are rare cases where refs could cause problems, especially in code that both gets and sets a string.
- If the string is unset, then GetS returns a newly constructed empty string. The alternative would be to extern declare a global empty string and return a const reference to it.

What do you think?

src/globals.h
ds-sloth: Are you saying that this is the proper style to construct a new empty string? (I don't think we'd be allowed to return a reference to a stack-constructed string, whether it's constructed `std::string` or by casting `""`.)

Wohlstand: Various static analyzers blame this case as it may. 

I meant, const reference, it bumps the life of temporary object, so:

```cpp
void call1(const std::string &one) {}
void call2(std::string &one) {}

call1(std::string()); // will work fine
call2(std::string()); // will fail at compilation
```

src/blocks.cpp
```diff
@@ -1463,34 +1473,45 @@ void KillBlock(int A, bool Splode)
...
-                    tempBool = true;
-            }
-
-            for(C = 1; C <= numBlock; C++)
-            {
-                if(C != A)
+                for(int npc : Layer[C].NPCs)
                 {
-                    if(Block[A].Layer == Block[C].Layer)
+                    if(NPC[npc].Generator == false)
```
Wohlstand, line 1499: damn Redigit
```cpp
if(!NPC[npc].Generator)
```

src/globals.h
ds-sloth: Interesting, okay, then I will make this change and make it a const ref. Thanks for the tip!

ds-sloth: Wait, this isn't the case I'm dealing with here. You can't return a reference to a stack-allocated std::string() even if the reference is const, because the reference will be deallocated as soon as the function returns. I'll make a global const empty std::string instead.

src/globals.h
Wohlstand: I do mainly worry about the Autocode side which goes to use this widely (I want to make Autocode class hold string indexes and AutocodeManager operate with them too, and so, if Autocode wants to print any string, it will give the index, and it will be printed every frame, so, I want it should use the const reference to don't copy the string when it's not needed)

src/globals.h
Wohlstand: I see. Should be fine then.

src/graphics/gfx_credits.cpp
```diff
@@ -90,7 +90,7 @@ void DrawCredits()
             break; // Nothing also to draw
 
         // Printing lines of credits
-        SuperPrint(c.Text,
+        SuperPrint(GetS(c.Text),
```
Wohlstand, line 93: In a case like this, REALLY wants GetS() to use the const reference to avoid copyings of string as much as possible.

src/graphics/gfx_update.cpp
```diff
@@ -2630,9 +2631,9 @@ void UpdateGraphics(bool skipRepaint)
 //                    BitBlt myBackBuffer, .X - vScreen(Z).Left, .Y - vScreen(Z).Top, 32, 32, GFX.ECursor(2).hdc, 0, 0, vbSrcPaint
                     XRender::renderTexture(curX, curY, GFX.ECursor[2]);
 //                    If LCase(frmLayers.lstLayer.List(frmLayers.lstLayer.ListIndex)) <> "default" Then
-                    if(!e.Layer.empty() && SDL_strcasecmp(e.Layer.c_str(), "Default") != 0)
+                    if(e.Layer != LAYER_NONE && e.Layer != LAYER_DEFAULT)
 //                        SuperPrint UCase(frmLayers.lstLayer.List(frmLayers.lstLayer.ListIndex)), 3, .X + 28, .Y + 34
-                        SuperPrint(EditorCursor.Layer, 3, curX + 28 , curY + 34);
+                        SuperPrint(Layer[e.Layer].Name, 3, X + 28, Y + 34);
```
Wohlstand, line 2636: Are you sure it's valid X and Y instead of `curX` and `curY`? Isn't a mistake?

src/graphics/gfx_update.cpp
```diff
@@ -2609,9 +2610,9 @@ void UpdateGraphics(bool skipRepaint)
 //                    BitBlt myBackBuffer, .X - vScreen(Z).Left, .Y - vScreen(Z).Top, 32, 32, GFX.ECursor(2).hdc, 0, 0, vbSrcPaint
                     XRender::renderTexture(curX, curY, GFX.ECursor[2]);
 //                    If LCase(frmLayers.lstLayer.List(frmLayers.lstLayer.ListIndex)) <> "default" Then
-                    if(!e.Layer.empty() && SDL_strcasecmp(e.Layer.c_str(), "Default") != 0)
+                    if(e.Layer != LAYER_NONE && e.Layer != LAYER_DEFAULT)
 //                        SuperPrint UCase(frmLayers.lstLayer.List(frmLayers.lstLayer.ListIndex)), 3, .X + 28, .Y + 34
-                        SuperPrint(e.Layer, 3, X + 28, Y + 34);
+                        SuperPrint(Layer[e.Layer].Name, 3, X + 28, Y + 34);
```
Wohlstand, line 2615: The same as on the world side, is that valid to use the X and Y instead of `curX` and `curY`? At this line, looks like my own bug because I did this mess. Could you verify?

src/layers.cpp
```diff
@@ -793,50 +1189,55 @@ void UpdateLayers()
...
                 if(g_compatibility.enable_climb_bgo_layer_move)
                 {
-                    int allBGOs = numBackground + numLocked;
-                    for(B = 1; B <= allBGOs; B++)
+                    for(int B : Layer[A].BGOs)
                     {
-                        if(BackgroundFence[Background[B].Type] && Background[B].Layer == Layer[A].Name)
+                        if(BackgroundFence[Background[B].Type])
                         {
+                            // @Wohlstand, should this be 0?
```
Wohlstand, line 1196: No, this is required for player to inherit the move speed to move together with fences

src/npc/npc_update.cpp
```diff
@@ -5159,6 +5178,8 @@ void UpdateNPCs()
                                         }
                                     }
                                 }
+
+                                // @Wohlstand, should we add something here for the BGO fence fix?
```
Wohlstand, line 5182: BGO fence fix doesn't require any structure side fixes, it needs the fix at the player logic and at layer processing logic, which was done already.

src/layers.cpp
ds-sloth: But this is the section where the layer is stopped (see above `Block[B].SpeedX = 0`)

Wohlstand:
I checked the rest of the code, and the thing you can complete: do use string indexes at `Autocode` class freely (MyRef and MyString), but, be aware of specific commands from the enum which expects the layer or event name, so, at the `AutocodeManager` where all `Autocode` things got constructed, do select generic string, layer name, or event name depending on the luna-command from the enum to assign the index of `MyString`. where the `MyRef` is used, do replace strings at my index tables with stringindex to use it instead of raw strings.

src/graphics/gfx_update.cpp
ds-sloth: I haven't backported the new editor drawing code yet, but I'll change these to `curX` and `curY` for now. The new editor drawing code also used `curX` and `curY` here.

src/layers.cpp
Wohlstand: I guess, then, should be right, but please, test these demos to make sure nothing got broken:
[fence-tests.zip](https://github.com/Wohlstand/TheXTech/files/7972644/fence-tests.zip)
You may toggle between modern and vanilla modes to see the behavior differences

src/graphics/gfx_update.cpp
ds-sloth: You are right, thank you!

src/npc/npc_update.cpp
ds-sloth: The problem is that this could allow the `SpeedX` of the BGOs to remain non-zero forever, since there is nowhere else to set the `SpeedX` of the BGOs.

ds-sloth:
I think that I have already updated autocode to use Event indices instead of names. It seems that it did not use Layer names at all originally.

Is this the behavior you expect on the test level?

Scr_2022-01-31_12-41-54

src/globals.h
ds-sloth: I don't think that Autocode uses the level strings right now, but uses its own strings instead. But I have updated it to use const refs now.

src/layers.cpp
ds-sloth: I will think more about these later -- we may also need to add extra test cases (starting and stopping the layer's motion, attaching the layer to an NPC, etc) to test all of the unusual cases I'm discussing here. Let's not deal with that during this merge.

src/npc/npc_update.cpp
ds-sloth: But here, this is a case where the NPC is setting the layer's speed, so I'm not sure if we need to do anything here. Let's talk about this separately.

Wohlstand:
Looks like yes:
- 1 stops all music and show germeni meme
- 2 start Escape music and shown alphabet picture
- 3 starts World Bowser music 
- 4 starts loop of triggering events
- 5 breaks loop of 4 and show Rosalina picture

ds-sloth:
Great. Then I think we are fine to merge if you think so too. These changes should be familiar to Axiom also since they have been the system at the devel branch for a while.

Wohlstand:
Btw, what I meant Autocode thing is still use raw strings, I meant this:
https://github.com/Wohlstand/TheXTech/blob/ce0587a6bdb0d8266cdc154deb41120f855e5f68/src/script/luna/autocode.h#L247-L248
they can be freely replaced with indexes system

ds-sloth:
>  they can be freely replaced with indexes system

I will add this in a minute. Anything else you think we should do before merge?

Wohlstand:
> I will add this in a minute. Anything else you think we should do before merge?

Would be better, if you can, but if you want to take a break, then let's after merge

ds-sloth:
It was good to do this -- it also made me reorganize, moving the string-specific globals into a new `global_strings.h` to reduce the huge `globals.h` mess. Will upload in a minute after merging.

ds-sloth:
I've updated the changelog and I am ready to merge if you are.

Wohlstand:
In the evening I'll take another look, right now I am busy with my main job.

ds-sloth:
Let me know if you are ready to merge tonight. It will be helpful for me so I can begin porting some additional features (first, probably, the new editor, since it does not interfere with the gameplay code much).

Wohlstand:
I do have a question against the `Used P Switch` layer: Are you really sure no levels in a wild that has such layer name in their list? I rely on an exact name for the new-added layer name, it's a chance that one of the old levels would be broken. The case is absolutely similar to the speed-run timer that got stopped by the "Boss dead" event name that got triggered by mid-boss at The invasion 2.

PR #268, opened 2022-01-31
Added custom memory allocator for RenderOp to optimize the real-time performance

Wohlstand:
Using dynamic allocation/releasing every frame is not good as the default generic-purpose allocator can affect the performance and stability. So, to keep the flexibility of dynamic allocation and improve the performance, it's reasonable to use the custom allocator for the stuff.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -169,6 +169,7 @@ if(USE_STATIC_LIBC)
     set(BUILD_SHARED_LIBS OFF)
 endif()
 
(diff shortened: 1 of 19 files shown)


ds-sloth:
As long as this only modifies the autocode sections, it doesn't make a difference to me. Feel free to merge whenever you feel it is polished.

Wohlstand:
The thing is very small and for me, looks just fine. So, thanks for a look :fox_face: :+1:

PR #299, opened 2022-02-28
In-game editor

ds-sloth:
The in-game editor has been ported and is ready for testing and feedback as we prepare to merge. To enable the editor for testing, please set the "enable-editor" flag in the "[main]" section of the settings ini. Additionally, the editor icons found here (https://github.com/Wohlstand/TheXTech/blob/master/docs/EditorIcons.png) should be placed in the ui asset directory.

It should currently have full game-controller support (without multiplayer / multicursor at this time) for in-couch editing and testing.

It currently lacks support for most TheXTech-exclusive features, but it will not damage levels that include them.

In case of any issues, please back up any of your own work before editing it in the in-game editor at this time.

On my side, I think we are ready for a merge, maybe with a minor notice that the editor is in a beta state. However, I am sure that community members will have some good feedback before we do so.

Todo list:
- [x] Confirm and fix controls issues
- [x] (need test) tested by @0lhi , thank you!
- [x] Implement the new npc_special_data system
- [x] (need Wohlstand's feedback and test)
- [x] Implement all (most) of Wohlstand’s other new features
- [x] Tune the controls for devices without two directional controls


0lhi:
I cannot navigate the levels with arrow keys unless I start testing and leave it again.

I can control Mario with a keyboard, but not with a controller.

When I repeatedly press Start with a Controller, this happens:

Scr_2022-02-28_18-57-27

(Start doesn't leave the Pause Menu, but seems to "reopen" it.)

src/editor/write_level.cpp
```diff
@@ -0,0 +1,406 @@
...
+
+        if(n.Type == 91 || n.Type == 96 || n.Type == 283 || n.Type == 284)
+            npc.contents = n.Special;
+
+        if(n.Type == 288 || n.Type == 289 || (n.Type == 91 && n.Special == 288))
+            npc.special_data = n.Special2;
+        else if(n.Type == 260 || NPCIsAParaTroopa[n.Type] || NPCIsCheep[n.Type])
+            npc.special_data = n.Special;
+        else if(n.Type == 86)
+            npc.special_data = n.Special7;
```
Wohlstand, line 194: Since the time, there are more NPCs got the support for Special7 such as rail platforms, bill blasters, etc.

ChristianSilvermoon:
So as mentioned over on Discord, I'll keep this comments up to date with my suggestions, findings, etc.

**Last Updated:** 2022/03/04 @ 1:35 AM (UTC)

### Polish Suggestions

#### Improve UI Visuals
* Try replacing the top bar's flat color with a 4 tone gradient to fit with the 16-bit aesthetic and add a simple border-frame (the sides of the frame can/should be cropped on lower resolutions
image
* For the large menu, use the same transparent black background seen in the Abstract Controls Menus and extend the frame down/around it
* Add a splash of color to some more of the editor icons 
image
* For the Music Setting in the menu, use the Music Note Editor Icon as opposed to the Orange Hand.
* For True/False checkboxes in the UI, use the "✔" or "✘" Editor Icons instead of a green outline
* For the Destroyed Blocks Layer, use the "block-4" graphic as opposed to the orange hand
* For the Spawned NPCs layer, use the "npc-1" graphic as opposed to the orange hand

#### Visual Feedback
* When switching sections using Editor hotkeys display "Section 2", "Section 3", etc in the middle of the screen for a few seconds.
* Display text centered below the top bar that identifies the hovered element (IE "Selection Tool", "Eraser", "Blocks", "Background Objects", "NPCs", "Water and Quicksand", "Warps", "Level and Section Settings", "Layers", "Events", "Save and Load", "Play Test")
* When a layer other than Default is selected, display the layer name next to the cursor (As in the original SMBX editor)
* If possible add an option to dim the graphics that belong to layers other than the one selected.
 
#### Audio Feedback
* When clicking UI Elements, play "sound-74" (`smw-saw.ogg`)
* When changing sections, play "sound-26" (`slide.ogg`)
* When toggling large menu, play "sound-30"  (`pause.ogg`)

#### Quality of Life
* During Playtesting, bring up the pause menu on death.
  * This can prevent being stuck in death loops if you die too quickly after re-spawning. It can be very difficult to pause in time in these situations and this would effectively solve that problem for the play tester.
  * Can be made optional with a 'editor-pause-on-death' setting

#### Bugs Found
* Right joystick behaves strangely
* Cursor does not line up with where clicks occur on the on screen keyboard in Editor
image

P-Star7:
Why are some of the SMW 16x16 hill sceneries enlarged, and others are not?

Also don't worry, if the editor UI is in the repo then I can enhance the icons later.

ds-sloth:
@P-Star7 to my knowledge they are all treated the same way, maybe the hills that are not enlarged are actually on 32x32 canvases?

src/editor/write_level.cpp
ds-sloth: Should we just set `npc.special_data = n.Special7` for *everything*?

ds-sloth:
@ChristianSilvermoon I fixed the cursor line-up issue, thanks for reporting it.

> When a layer other than Default is selected, display the layer name next to the cursor (As in the original SMBX editor)

This should already be done. Are you not seeing it?

@Wohlstand I should have fixed the MagicHand integration, but I realize, there are certain events at the PGE editor that might break the integration (making a new layer, mainly). Should we worry about this or keep it as-is?
Similarly, any changes in the TheXTech editor (like changing a block's ID after picking it up) are not reflected. Should we try to fix this, or should we disable the in-game editor UI when Interprocess is active?
Also, I am surprised that Water/Quicksand can't be changed during MagicHand, and I might change this at the in-game editor. If you want to synchronize with this, you would need to update the Moondust editor on your side also.

ChristianSilvermoon:
> @ChristianSilvermoon I fixed the cursor line-up issue, thanks for reporting it.
> 
> > When a layer other than Default is selected, display the layer name next to the cursor (As in the original SMBX editor)
> 
> This should already be done. Are you not seeing it?
> 
> @Wohlstand I should have fixed the MagicHand integration, but I realize, there are certain events at the PGE editor that might break the integration (making a new layer, mainly). Should we worry about this or keep it as-is? Similarly, any changes in the TheXTech editor (like changing a block's ID after picking it up) are not reflected. Should we try to fix this, or should we disable the in-game editor UI when Interprocess is active? Also, I am surprised that Water/Quicksand can't be changed during MagicHand, and I might change this at the in-game editor. If you want to synchronize with this, you would need to update the Moondust editor on your side also.

I was not, but I'll check and report back when I can

Wohlstand:
> PGE editor

No longer "PGE", there is "Moondust"

> making a new layer, mainly

Making a new layer usually makes no significant sense, but probably, you can drop an unknown layer name into just "Default". Magic hand functionality is intended to support the next actions:
- place a new block / BGO / NPC
- receive property change for a floating item
- pick-up items from the game scene and send back current properties to get the ability to edit them at Editor side
- erase items using a rubber tool (can be called using a middle mouse button)

src/editor/write_level.cpp
Wohlstand: :thinking: 
Not so, just for these NPC that got this rule at the side of the main level file loader. There are some custom rules that could be done.

Wohlstand:
>  Similarly, any changes in the TheXTech editor (like changing a block's ID after picking it up) are not reflected. Should we try to fix this, or should we disable the in-game editor UI when Interprocess is active?

When the interprocess mode is active, usually it has priority because the external editor takes the rest of the controlling

Wohlstand:
>  Also, I am surprised that Water/Quicksand can't be changed during MagicHand, and I might change this at the in-game editor. If you want to synchronize with this, you would need to update the Moondust editor on your side also.

Vanilla SMBX does the same, it blocks an ability to modify PEZ (Physical Environment Zone) and warps

ds-sloth:
> you can drop an unknown layer name into just "Default".

This is what the system does currently.

> When the interprocess mode is active, usually it has priority

Sure. I just added code to disable the new editor UI during Interprocess MagicHand testing.

> Vanilla SMBX blocks PEZ and Warp modification

Okay, I will preserve this behavior.

ds-sloth:
@Wohlstand I'm not certain how much MagicHand is meant to do. If I place a block in TheXTech, should it show up in Moondust? If I place it in Moondust, should it show up in TheXTech? That's not happening right now.

Wohlstand:
> If I place a block in TheXTech, should it show up in Moondust? If I place it in Moondust, should it show up in TheXTech? That's not happening right now.

No, items placed/erased at Game were don't affect the in-Editor scene, the in-game editing is a testing sandbox and not actual editing, the same is going at Vanilla SMBX. All fine :wink: 
The only thing if you picking-up an item, Editor should pop up the properties toolbox to modify the settings of a floating object.

Wohlstand:
So, the main way of Magic hand use is making quick testing at the running instance, to place/erase random NPC or objects and play with them, basically, could serve as a way to quickly debug behavior of a new-crafted NPC or something also done with scripts

ds-sloth:
@ChristianSilvermoon Whenever you have time please test the new version. I believe I’ve done everything you suggested that I’m planning on implementing, except for fixing the controls issues. Thanks a lot for your comment, it was really helpful.

ds-sloth:
I won't be able to work on the project as much for the few days or week. I'm responding to each of your unchecked suggestions here:

**Polish**:

> For the large menu, use the same transparent black background seen in the Abstract Controls Menus and extend the frame down/around it.

I like the pale blue color I picked. I can see what happens if we make it transparent but I'm not sure it would be good. We could also give the editor 2 themes perhaps.

> Consider removing the dark squares behind icons.

They're there to indicate where the hitboxes are and to keep the UI consistent. I could slightly change their color if that would address your complaint?

> For the Music Setting in the menu, use the Music Note Editor Icon as opposed to the Orange Hand.

In my design language, the Orange Hand consistently indicates any action that will take the user to a new screen, and the Orange Pencil indicates any action that will open the text editor screen. I'd rather not change this here. If you think there's a better icon to use to indicate subscreens, we can change the Orange Hand to that, but I'd rather keep it consistent.

> For True/False checkboxes in the UI, use the "heavy_check_mark" or "✘" Editor Icons instead of a green outline

We could do this, but I'm not sure whether an empty box or the X icon is better. I don't want to make the UI too noisy. In either case, this would be in addition to the "selected" indicator, not instead of it.

> Destroyed Blocks and Spawned NPCs layer graphics

I consistently use the White Hand for any selections in lists. If you feel strongly about this I would be willing to reconsider.

**Feature requests**:

I also like your layer fading indicator and your limited input device suggestions. There are technical challenges that make implementing either of them nontrivial.

**Bug reports**:

Thanks for these. I will be able to figure each of them out except for the screen fade on exit. I'm not familiar with how @Wohlstand has implemented the screen fade, and especially uncertain how he prolongs the previous screen for a few frames during the fadeout. Wohlstand, if you could explain how I'd implement this fadeout, I can code it up.

ChristianSilvermoon:
> I like the pale blue color I picked. I can see what happens if we make it transparent but I'm not sure it would be good. We could also give the editor 2 themes perhaps.

The pale blue color is good, I like it. Supporting two themes sounds like a good option too. Transparent pale blue would probably be better than the black, especially if the layer fade thing is ever added.

> They're there to indicate where the hitboxes are and to keep the UI consistent. I could slightly change their color if that would address your complaint?

My main issue is that they look a little odd, but I understand why they're there. They stand out in a way that makes it look sort of tacked on though. But I do think it's a good idea to give an indicator of the hitboxes, particularly for smaller items. Maybe some way to make them blend more with the UI? Not entirely sure. :thinking:
 
> In my design language, the Orange Hand consistently indicates any action that will take the user to a new screen, and the Orange Pencil indicates any action that will open the text editor screen. I'd rather not change this here. If you think there's a better icon to use to indicate subscreens, we can change the Orange Hand to that, but I'd rather keep it consistent.

I get what you mean, and that actually makes a lot of sense, but yeah perhaps some other icon that's more fitting, though not sure what at the moment. :thinking:
 
> We could do this, but I'm not sure whether an empty box or the X icon is better. I don't want to make the UI too noisy. In either case, this would be in addition to the "selected" indicator, not instead of it.

Empty box for unchecked is better, unless it could be confused with another UI element.
 
> I consistently use the White Hand for any selections in lists. If you feel strongly about this I would be willing to reconsider.

Perhaps some other icon that feels more indicative. Usage of the hand like this makes it look like placeholder graphic. Much like orange hand not sure what to suggest at the moment though.

> I also like your layer fading indicator and your limited input device suggestions. There are technical challenges that make implementing either of them nontrivial.

Very understandable, they're bigger things.

> Thanks for these.

You're welcome, and thank you so much for all you've put into this, you're awesome ^^

ds-sloth:
Hi @ChristianSilvermoon if you have a chance in the next week or so to give the editor another pass, I've polished a lot of the points you suggested and should have fixed all of the bugs. I've also implemented most of the TheXTech-specific options (including some warp transition effects that might not be implemented yet?).

Next time I have time to work on it I'll probably work on the limited controls setup or layer highlighting feature.

ChristianSilvermoon:
Support for Portal Warps Appears to be missing, but otherwise, looking and feeling pretty great ^^

Scr_2022-03-29_00-50-58
Everything here was made using only the in-game editor.


There's the instant warp, but that doesn't conserve your momentum... 

PS: Also seems I bypass this warp moving upwards too fast from the canon, but *do* trigger it on the way back down, but that's likely more of a collision checking issue than anything to do with the editor :P

ds-sloth:
@ChristianSilvermoon to follow up with what we discussed on Discord yesterday, here is a page for editing the Test Play settings. I also tuned the post-death pause screen and made sure that Magic Hand worked properly with screen shake.

Will push the commits momentarily.

ds-sloth:
@Wohlstand please review the code at https://github.com/Wohlstand/TheXTech/blob/wip-new-editor/src/npc_special_data.h and let me know if this is an acceptable way to track legacy NPC behaviors. I integrated it into the level loading procedure so that we do not need to include special cases. Since everything here is inline functions on constexpr data tables, the linear searches in `find_legacy_Special7` should be optimized into a fast set of conditionals, so I do not think that calling them on every NPC loaded will cause a major slowdown.

ChristianSilvermoon:
I interestingly just had this happen while returning to the title screen :P
image
the text seems like it's just kinda stuck

ds-sloth:
Ha, that was unexpected. Thanks for the report! This may be a general problem where the quick reconnect UI doesn’t disappear on the menu screen. I bet I’ve got the render code plugged in but am missing the logic somehow. I’ll take a look and fix this before the merge, though I suspect it’s a general abstract controls issue.

ds-sloth:
Fixed the bug you reported (thank you!) and also a few other minor bugs that I found in the Drop/Add screen during Level Test mode.

LoveBodhi:
This editor is much like Mario Editor, but that's one is complex.

LoveBodhi:
The only remaining is Touch Screen.

ds-sloth:
Since Wohlstand is busy with other work, I will try to implement some touchscreen code for the New Editor before he has time to review on Saturday.

LoveBodhi:
I still need to verify the new editor after finish.

ds-sloth:
Powered up eraser and added the single vs multi-type ("axe mode") erase distinction as we discussed in Discord.

LoveBodhi:
Axe mode is like Mario Editor, when clear the selection, will remove all objects.

LoveBodhi:
Found some problems

Scr_2022-04-23_16-45-38

Sizable block can only be placed one per ID

Scr_2022-04-23_16-32-41

Custom GFX effect vanish when enter world name

Scr_2022-04-23_16-33-53

Scr_2022-04-23_16-35-54

Custom / Extended Slot doesn't shown, and the music name displaying issue

Scr_2022-04-23_16-50-18

Doesn't support custom block GFX

(I used Nostalgic Paradise for testing)

LoveBodhi (changes requested):
If there's asset-based elements and custom block GFX, that's fine.

ds-sloth:
Thanks @LoveBodhi for the review!

I've fixed the issue you spotted where the title's assets were unloaded. Thank you for reporting this.

Some questions:

> Sizable block can only be placed one per ID

I can't reproduce this issue. Can you follow up with a screenshot including you selecting the sizable block you are trying to place?

> Custom / Extended Slot doesn't shown,

The in-game editor is not intended to show options that are absent from the SMBX editor with only a few exceptions. Can you explain how you would want this to be changed? I can consider this feature but I do not think it is needed before we merge the in-game editor.

> and the music name displaying issue

What is the issue here? How would you want the music names to be displayed?

> Doesn't support custom size of block GFX

Is this feature supported in SMBX? Where is custom block GFX size set?

LoveBodhi:
> What is the issue here? How would you want the music names to be displayed?

Asset-varied

LoveBodhi:
Also, Tileset Organization should be asset-varied instead of just SMBX 1.3 constant.

ds-sloth:
The infrastructure for this option does not exist at the moment, but I can add an option to display them in ID order without headings if that would be useful?

LoveBodhi:
> The infrastructure for this option does not exist at the moment, but I can add an option to display them in ID order without headings if that would be useful?

That should be...

src/control_types.h
```diff
@@ -51,21 +51,22 @@ struct CursorControls_t
...
     bool ScrollUp = false;
     bool ScrollDown = false;
+
     // call to easily send cursor offscreen (on disconnect, etc)
     inline void GoOffscreen()
     {
-        X = -128.;
-        Y = -128.;
+        X = -4000.0;
+        Y = -4000.0;
```
Wohlstand, line 59: Why not to make the sort of -9999999999999 (something ultra-huge and being constant in binary form to avoid possible false positives, or something, but right, Editor won't allow non-multiple 32 camera location being set).

src/graphics/gfx_update.cpp
```diff
@@ -601,6 +598,8 @@ void UpdateGraphics(bool skipRepaint)
                 GetvScreenCredits();
             else
                 GetvScreen(Z);
+        }
+        {
```
Wohlstand, line 602: 纳尼？ :eyes:

Should be here an "else" branch, or please add the splace between these brices blocks to don't confuse...

Wohlstand:
Okay, I had checked the most of basic stuff on the PC platform, and seems it's fine if don't count some places which need to be tested/polished more. That could be done later, and Editor itself could be kept as an "experimental" thing even after release.

Wohlstand (approved):
All seems to look fine, we can merge this soon if nothing special is needed. All bug fixes related to the Editor could be done after merging easily. Also, I will need to test out the work of things on Android using the touchscreen and plugged gamepad.

The editor seems still needs more polishing (regarding events settings, etc.), and so, even if we release the general thing, we can mark this editor as an experimental thing easily.

Wohlstand:
Keeping in a mind, the menu bug still not been fixed, so, keeping it to be fixed later into branch "master" directly.

PR #304, opened 2022-03-02
Polishing the global strings system

Wohlstand:
- Unique strings will be kept as the same indices
- Made sure if assign the same string field with another value, the new string entry will be allocated as a new unique string added
- Every known string will be tracked with usages counter and will be kept for the future use

This whole work should resolve the crash while running certain LunaDll scripts that have to add/remove code entries often: because old and duplicated strings get never removed, the whole bank gets filled with duplicates because of an infinite cycle of creating/removing string entries somewhere.

The test level that should work correctly here:
[ThroughMetro.zip](https://github.com/Wohlstand/TheXTech/files/8173468/ThroughMetro.zip)
Without changes from this PR, the whole bank of strings gets filled and crashed because of the assert failed

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -325,6 +325,7 @@ set(THEXTECH_SRC
     src/sound.cpp
     src/frame_timer.cpp
     src/global_dirs.cpp
+    src/global_strings.cpp
     src/core/base/render_base.cpp
     src/core/base/window_base.cpp
     src/core/base/msgbox_base.cpp
--- a/src/global_strings.cpp
+++ b/src/global_strings.cpp
@@ -1,0 +1,145 @@
+/*
+ * TheXTech - A platform game engine ported from old source code for VB6
+ *
+ * Copyright (c) 2009-2011 Andrew Spinks, original VB6 code
+ * Copyright (c) 2020-2022 Vitaly Novichkov <admin@wohlnet.ru>
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "global_strings.h"
+
+
+// utilities for stringindex_t
+const std::string g_emptyString = "";
+//! Strings storage
+static std::vector<std::string> g_LevelString;
+//! Counter of usages per every string
+static std::vector<int> g_LevelStringUsages;
+//! List of free indexes appearing at the moddle of array
+static std::vector<int> g_FreeIndexes;
+//! String-To-Index map
+static std::unordered_map<std::string, int> g_uniqueStringsIds;
+//! Number of world strings
+static size_t g_numWorldString = 0;
+
+static std::vector<std::string> g_LevelString_backup;
+static std::vector<int> g_LevelStringUsages_backup;
+static std::vector<int> g_FreeIndexes_backup;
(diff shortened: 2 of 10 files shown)


ds-sloth:
I agree that we need a better system if strings are being frequently deallocated. Some minor questions about this solution:

(1) PtrS is meant to return a pointer to a writable string, so you should ensure the index is unique when it is called.

(2) While the problem led to excessive RAM use, this doesn’t seem especially resource-friendly solution. (Maintaining multiple copies of the strings, new vectors of ints, map checks, etc.)

If you asked me for my approach, I would probably propose using unique string indexes per object (allowing one string to appear in multiple places) but adding a FreeS method for when an object with a stringindex member is destroyed, which would add its stringindex to a LIFO free index stack. Does that seem like a reasonable solution to you?

Wohlstand:
> (1) PtrS is meant to return a pointer to a writable string, so you should ensure the index is unique when it is called.

Right now (at least in the "master" branch), nobody uses this call, and then, the case will need to be checked

The frequent use of SetS goes at the "Autocode" engine right now (if mean runtime, not load time), and GetS is remains the same behavior as before of my rework. So, the most of code doesn't bother about the SetS thing. However, speaking about making copies of objects, I can replace the `SetS(target, GetS(source))` with just `target=source;` and so, no bothering on this

Wohlstand:
> Right now (at least in the "master" branch), nobody uses this call, and then, the case will need to be checked

@ds-sloth, about PtrS(), do you have exact examples with using of such pointer for write purpose? The hypothetical case is used in the editor to modify string fields anywhere, etc. But should be fine to use the SetS as it is not intended to be spammed for every frame. For reading purposes, the GetS works the same fast.

P.S. I fixed the Autocode side to simply copy the indexes than attempt to SetS(GetS) to avoid such spam I described

ds-sloth:
PtrS was implemented for use in the in game editor (part of why I ported strings before the editor). We can check on it when the editor port is merged.

So, I think there are three possible solutions:
(1) full reference counting, what you are implementing here
(2) unique string indices with a free stack, what I am proposing
(3) old solution, but just do target = source in cases where source is constant over the level, and target is a temporary object

(3) is the simplest but it wouldn’t work if autocode dynamically creates strings that belong to temporary objects.

(2) is very simple and resource cheap unless there are MANY duplicates of single strings (cheaper than 1 unless more than half of strings are duplicated)

One good thing: both (1) and (2) can use compatible calling conventions outside of the global strings module itself, so we can retune between the two of them in the future.

However, both (1) and (2) require us to use SetS(target, GetS(source)) to make target unique (2) or increment its reference count. Additionally, unless we implement FreeS to reduce the reference count of deallocated objects’ strings, then none of these solutions will be better than (3), since temporaries would never be freed.

Simple ‘target = source’ wouldn’t work for the reference counting you are implementing here since it prevents us from accurately counting references. Please don’t do that unless we are committing to option 3 (old simple code, where temporary objects don’t own their strings)

ds-sloth:
PtrS is used to set the target string of the file browser for a Warp’s target level when the file browser is *started*. It doesn’t finish until many frames later, might not succeed, etc, and the file browser doesn’t know whether it was called on a Warp, on a section’s custom music, on a world music’s custom music (new feature that would also use PtrS). So this case strongly benefits from having a freely writable string pointer. (See example here: https://github.com/Wohlstand/TheXTech/blob/f88c508b7ef89fa957aa435dc9e3dbacc670a4f2/src/editor/new_editor.cpp#L2996

I don’t think we have any string comparisons in our codebase, so I don’t think optimizing them is a huge goal right now, especially since we may prefer option (2) if we are frequently setting strings to dynamic values from autocode.

Wohlstand:
For the testing, I made two modes of work:

- With uniqueness (new mode): there are total 76 entres created
Scr_2022-03-03_22-54-16

- Without uniqueness (old mode): there are total 28 entries created
Scr_2022-03-03_22-57-44

ds-sloth:
This is looking better! Can we rename the new single argument SetS to AllocS, and also add a new FreeS method?

Then on my side, I can update the version without uniqueness counting to support the FreeS command also.

Wohlstand:
> This is looking better! Can we rename the new single argument SetS to AllocS, and also add a new FreeS method?

Good idea!

Wait for the next commit

Wohlstand:
Done

Wohlstand:
As we discussed at Discord, gonna merge this

PR #311, opened 2022-03-18
Make speedrun timer be more colorful

Wohlstand:
This update makes the speed-run timer being more colorful when it stops at the level end or episode complete events.

Scr_2022-03-18_03-16-11
Scr_2022-03-18_03-13-54

Any ideas/suggestions/objections?

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -19,6 +19,7 @@ Changelog for 1.3.6 [WIP]
 -Fixed the vanilla bug: a Player can stuck in the pipe when enters it being a fairy (@Wohlstand)
 -Added an option to enable the fast walking on the world map (@ds-sloth, (@Wohlstand)
 -Added an ability to customize UI/common assets from the episode and level sides (@Wohlstand)
+-Added a blink effect at speedrun timer on level/episode completion (@Wohlstand)
 
 Changes for 1.3.5.3-1
 -Fixed autostart events behaviour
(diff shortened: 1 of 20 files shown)


Wohlstand:
> Coloring/Blinking should be optional via an [effects] or [speedrun] entry in settings.ini/compat.ini: enable-timer-effects. It should have 3 options:

I guess, there is need the fourth `undefined` value and the priority:
- compat.ini by default sets one of values (enable/disable/opaque) by default
- settings/thextech.ini sets the "unspecified" by default, meaning, episode-side config is preferred
- when settings/thextech.ini defines anything also than "unspecified", it will have highest priority over compat.ini, means user has different wish against the look

ds-sloth:
Ah, this is a great effect! Very simple, really nice to look at. I agree with @0lhi that it's probably a good idea to enable by default only if the user chose opaque speedrun timer.

> I guess, there is need the fourth undefined value and the priority

After we merge multires + in-game editor, my next goal is the big compat / config rewrite we have been talking about so that we don't need to think about these anymore. :)

The code is clean and simple. No reason not to merge it in my opinion. :100:

0lhi:
*true*, *false*, and *opaque* settings all work as intended.

Didn't bother to test with compat.ini, cause imo we should wait until the big rewrite before encouraging it's use.

Wohlstand:
> Didn't bother to test with compat.ini, cause imo we should wait until the big rewrite before encouraging it's use.

compat.ini-side option will work when the settings/thextech.ini side is set into `undefined`

Wohlstand:
Going to merge this as seems, all fine now.

PR #315, opened 2022-03-28
Added UI customization by episode/level side

Wohlstand:
A very small update that adds an ability to customize UI assets from a level and/or episode side as regular CGFX.

Note: there are internal names being used to match files at UI directory rather than [repeating the list from this table](https://wohlsoft.ru/pgewiki/Hardcoded_Image_List_of_SMBX_Engine) which should be simpler for users than letting them think which thing for which, etc...

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -18,6 +18,7 @@ Changelog for 1.3.6 [WIP]
 -Fixed unexpected music got started after the level got entered too quick from the world map (@Wohlstand)
 -Fixed the vanilla bug: a Player can stuck in the pipe when enters it being a fairy (@Wohlstand)
 -Added an option to enable the fast walking on the world map (@ds-sloth, (@Wohlstand)
+-Added an ability to customize UI/common assets from the episode and level sides (@Wohlstand)
 
 Changes for 1.3.5.3-1
 -Fixed autostart events behaviour
--- a/src/gfx.cpp
+++ b/src/gfx.cpp
@@ -44,10 +44,7 @@ void GFX_t::loadImage(StdPicture &img, const std::string &path)
 }
 
 GFX_t::GFX_t() noexcept
-{
-    if(m_isCustom)
-        SDL_free(m_isCustom);
-}
+{}
 
 bool GFX_t::load()
 {
@@ -116,10 +113,7 @@ bool GFX_t::load()
         return false;
     }
 
-    if(m_isCustom)
-        SDL_free(m_isCustom);
-
-    m_isCustom = (bool*)SDL_malloc(m_loadedImages.size() * sizeof(bool));
+    SDL_assert_release(m_loadedImages.size() <= m_isCustomVolume);
     SDL_memset(m_isCustom, 0, sizeof(m_loadedImages.size() * sizeof(bool)));
 
     return true;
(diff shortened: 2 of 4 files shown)


LoveBodhi (approved):
OK. This is the good idea.

ajleite:
I’m happy to approve also, but I’d suggest you use C array of bools (or C++ bitarray) allocated in static rather memory instead of on the heap. And then add some static assert that checks that the array is the right size in case we add new UI assets.

ds-sloth:
I’m happy to approve also but I’d suggest that you replace the deque with a C++ bitarray or C array of bools with size specified at compile time and checked with a static assert. This keeps the memory statically allocated which is easier to track as we target low memory devices (and the bitarray is extra memory efficient).

Wohlstand:
> I’m happy to approve also but I’d suggest that you replace the deque with a C++ bitarray or C array of bools with size specified at compile time and checked with a static assert. This keeps the memory statically allocated which is easier to track as we target low memory devices (and the bitarray is extra memory efficient).

I used `deque` because `std::vector<bool>` is weird and breaks an ability to get a reference to each field which is needed for me. Anyway, a good idea, I could allocate it after, will try after sleep.

Wohlstand:
Tweaked just now

suicvne:
Looks like a straight-forward modification. I like it. @ds-sloth I also like your suggestion to replace `deque` with a standard C-array of bools. 

My only question, let's say for example BMVs.png gets overriden in one episode. By the time said episode is loaded, the "stock" BMVs.png is already loaded into memory at GFX.BMVs. Is the already loaded GFX.BMVs cleared/unloaded before the replacement is loaded? If it isn't, does the logic simply load the BMVs graphic again despite it technically already being loaded once? Would this cause issues on low memory platforms?

suicvne (approved):

Wohlstand:
> My only question, let's say for example BMVs.png gets overriden in one episode. By the time said episode is loaded, the "stock" BMVs.png is already loaded into memory at GFX.BMVs. Is the already loaded GFX.BMVs cleared/unloaded before the replacement is loaded? If it isn't, does the logic simply load the BMVs graphic again despite it technically already being loaded once? Would this cause issues on low memory platforms?

The algorithm is the same for all other CGFX:
- on loading a world map the first time, all level-specific CGFX including UI assets, gets loaded
- entering any level, cleans all level CGFX of the world map and loads new
- returning the world map back, it cleans all level CGFX and loads again at the world map
- etc.

- And yes, while CGFX is loaded, the original one is preserved in memory, once CGFX gets unloaded, the original one gets restored immediately

0lhi:
Seems good 👍

Scr_2022-03

ds-sloth:
So, I had suggested that you allocate your C array statically, but this is not a big difference. Feel free to merge. In the future I may try to optimize further for truly low-resource devices.

Wohlstand:
> So, I had suggested that you allocate your C array statically,

I made it dynamically to just simplify future adding of new items without pain, but wait, I can basically use an assert to crash when the array size is smaller than an actual number of elements... So, fool-proof protection can be done easily. :eyes:

ds-sloth:
The change looks clean; thanks for doing this!

Issue #190, opened 2021-09-21
Feature Request: Optional Backgrounds for Multires


0lhi:
When having a large resolution, 800x600 sections are rather unpleasant to look at:

![Scr_2021-09-21_15-46-35](https://user-images.githubusercontent.com/74732674/134186181-5820baed-ca5a-4e7d-9731-8611cfa51fea.png)

A solution for this would be to allow adding a custom background, for example as */graphics/ui/backgrounds/1280x720.png*.

Here's how it could look like using [this](https://pixabay.com/illustrations/background-circuit-grey-digital-2426328/):

![Scr_2021-09-21_15-46-35-Tech](https://user-images.githubusercontent.com/74732674/134187676-ae91e4f1-7307-4c53-b4c4-97a47c58efb9.png)

0lhi:
> but the game is drawing at 1440x720px (as it does on some Android phones).

Is there a dedicated 1440x720 setting on Android, or is that a result of the Dynamic setting?

I think, the dedicated settings should have two options: Using a Background that exactly fits the resolution (e.g. 1280x720.png), and using the one for the Dynamic setting as a fallback. Dynamic.png should be assumed to be a seamless repetition such as [this one](https://pixabay.com/illustrations/seamless-repeat-repetitive-2033661/).

The Game output should be outlined by the engine.

Making a beautifully framed image isn't really the point of this, but if some want to do so, they can just use Mode 2 or 3 to force an 800x600 output. Maybe that could be triggered by a Compat.ini setting, if there is demand for that.

The engine shouldn't do anything to the background aside from displaying it. Repeatedly in the case of Dynamic.png.

At least, that's what I think. Fancy stuff could be nice, but would be hard to implement, and I don't think that's worth it.

ChristianSilvermoon:
Suggestion: Base it on the same thing the World Map is doing currently, but maybe make it a different color? Maybe a dark color with the decorative shapes a slightly lighter shade?

0lhi:
The current World Map solution isn't reliable.

**_Dynamic SMBX:_**

![WorldMap](https://user-images.githubusercontent.com/74732674/134529589-a32a5df9-973d-4b59-86fd-b616fb88d4c7.png)

**_1280x720 Demo:_**

![WorldMap2](https://user-images.githubusercontent.com/74732674/134529595-6d19c58e-26a1-4530-aa55-5450b5f3ce79.png)

The Seamless approach would benefit it, too.

[...]

PR #328, opened 2022-05-08
Dynamic resolution support

ds-sloth:
I've had these features backported for a while. Now is a good time to begin testing and discussing, since they're now based on the main branch rather than another PR.

Some areas I think may have things to discuss:
- [x] Which NPCs should activate when they are visible, and which should activate when they are within 800x600 of the player? (Should the answer depend on whether the screen is bigger or smaller than 800x600?) Answer: we have a good set of defaults for now, future cases can be raised in Issues.
- [x] How should section resizes be animated? During 2P mode?
- [x] Should gameplay field resolutions above 1728x720 (2.4 aspect ratio 720p) be supported? Answer: probably not -- too far above the original game's 600p.
- [x] Correct handling to force at least 800x600 during Mode 2/3, and in multiplayer.
- [x] Finish the fog-of-war effect. Answer: still too many things to discuss, instead limit the world map view by default.
- [x] Ensure that the `ini` files for the new world map border and backdrop assets are correctly loaded.
- [x] How does this work on Vita? Other non-desktop platforms? Answer: @suicvne says there are a few issues with rendering that are more obvious here, but also occur in 2P mode on the non-multires version.
- [x] Resolve minor gameplay recording divergences recently discovered by @ds-sloth
- [x] Resolve 2P object spawn issues discovered by Sapphire Bullet Bill
- [x] Improve rendering for AutoCode HUDs that target 800x600. >800x600 done, <800x600 isn't possible in most cases. I now agree with @0lhi that it's up to the user's discretion if they're playing content designed for SMBX64/AutoCode.
- [x] Fix an issue where part of the menu information display became offscreen. Make menu logo transparent at low resolutions when it overlaps with text.

I'm sure there are also other suggestions or bugs that I haven't mentioned here. I'm looking forward to hearing lots of feedback.


ds-sloth:
There may be some frame-count compatibility issues with the qScreen code during 2 player mode. Need to investigate further. This code is really tricky to debug. However, we now have animations like this:

ChristianSilvermoon:
I do personally think higher resolutions should be supported, since 4K and 8K displays are a thing now days and it's always nice to be able to take advantage of the hardware you have.

Though admittedly I have a 4K screen I run at 1080p because that's just asking a bit too much of my poor 1050ti when it comes to gaming.

ds-sloth:
> I do personally think higher resolutions should be supported, since 4K and 8K displays are a thing now days and it's always nice to be able to take advantage of the hardware you have.

So, @Wohlstand built good scaling code and I've added integer scaling as an option. You'll certainly be able to run the game at 4K display resolution. (Even with integer scaling! It's 3x720p.) The question is whether the game's internal resolution should be able to be set above 1728x720. You can experiment by setting `InternalW` and `InternalH` in `config.ini` to something like 1920 and 1080 (or even 3840 and 2160) and seeing what things look like.

0lhi:
Could we have an option to have the window size fixed (except for dynamic), and automatically adjust to the res selection?

LoveBodhi:
Also widescreen.

ds-sloth:
@0lhi, I like your suggestion to resize the window automatically, but maybe with the following conditions: (1) the game is windowed, (2) the player has requested 0.5x, 1x, or 2x scaling, and (3) the player changes the playable field resolution to a non-dynamic setting. I had previously set it up to always resize the window, but certain strange issues kicked in on my Linux desktop, especially in fullscreen mode. I'll try to implement this feature again because I agree it's a lot more user friendly in certain contexts.

Separately, do you think we'll still need the cheat codes for changing resolution once there's an in-game options menu?

ds-sloth:
The window should now resize when changing settings -- thanks @0lhi for the suggestion!

0lhi:
> (1) the game is windowed, (2) the player has requested 0.5x, 1x, or 2x scaling, and (3) the player changes the playable field resolution to a non-dynamic setting.

I agree with all of these conditions. Though the third one doesn't seem to apply? The window still resizes, but the resolution isn't dynamic. If this is intentional, I suggest to hide the dynamic setting while an x-scale is selected and vice versa.

I also suggest to group the x-scalings together (like "0,5x - **1x** - 2x), and navigate through them with the select button. This prevents the window from becoming oversized when hitting 2x while having the 1280x720 resolution.

> Separately, do you think we'll still need the cheat codes for changing resolution once there's an in-game options menu?

Yep! Commands will always be quicker to access than menu settings. The easier to access they are, the more people will be encouraged to test levels.

ds-sloth:
Ah, I see. Yes, I'll make it not resize the window in this case. There are definitely use cases for mixing dynamic resolutions with fixed scaling, so I won't eliminate those combinations.

I understand that it's bad for the window to become resized, but the grouping mechanism with a new control key you mentioned seems a little overcomplicated. Maybe I'll try to figure out if SDL has any helpful methods for detecting the desktop size and use those to limit the requested window size.

Do you think the cheats should also resize the window when they are typed?

0lhi:
>  Maybe I'll try to figure out if SDL has any helpful methods for detecting the desktop size and use those to limit the requested window size.

Oh yeah, that would be ideal!

> Do you think the cheats should also resize the window when they are typed?

Can also depend on the scale option I think.

ds-sloth:
Is resizing during the game a cheat (should block save, etc)? (I really think not since this will be available from the in-game options menu, but wanted to check.)

0lhi:
> Is resizing during the game a cheat (should block save, etc)? (I really think not since this will be available from the in-game options menu, but wanted to check.)

Nah, this is a Debug Code :P

ds-sloth:
Alright, feel free to check these out whenever you have time!

0lhi:
I am very confused about the resize-ability of the window. Sometimes, I can't get it below 1280x720p. Sometimes, this occurs:
TinySMBX

Is it intended for the dynamic setting to not be able to go beneath 1280x720p?

Smallest Size at Integer Scale for GBA isn't supposed to be this big, is it?

IntegerScaleGBA

Idk. There is some pattern that can break the settings and some pattern that can fix them again. At least, scale1x + codes are reliable. I also like using the codes on nearest when the Window Size is fixed to 1280x720. It's convenient for the OBS Streaming software.

I like that the keyboard pops up when typing debugvie**w**, but the last letter finds it's way onto the keyboard:

Keyboard

Could you block out every non-number in that instance?

LoveBodhi:
Will be configable via gameinfo.ini?

ds-sloth:
@LoveBodhi, `gameinfo.ini` will be extended to set default settings and compat values that apply everywhere (unless overriden) but I think this will be implemented during the upcoming settings rewrite. At that point, we will be able to set the default internal dimensions game-wide, and also restrict the game from using larger play fields. Does this answer your question?

ds-sloth:
@0lhi, I think the weird resizability bug might have to do with the fact that we set the minimum window size whenever we set the window size. I don't know why we are doing this, but it's code I inherited from Wohlstand. I'll try removing it, and we can see if it breaks anything (maybe on Windows).

For the keyboard, this is a cheat -- it's not going to be all polished. I'll allow the user to type anything, even if it's invalid. But it's completely a bug that the last letter of the cheat shows up in the keyboard, and I'll try to fix that.

LoveBodhi:
> @LoveBodhi, `gameinfo.ini` will be extended to set default settings and compat values that apply everywhere (unless overriden) but I think this will be implemented during the upcoming settings rewrite. At that point, we will be able to set the default internal dimensions game-wide, and also restrict the game from using larger play fields. Does this answer your question?

I understand now, so Thanks❤

0lhi:
> I would keep the SDL_SetWindowMinimumSize() once to set the smallest size of the window (something ultra-small), just to avoid the user applying the impossibly small size.

240x170 seems reasonable.

240x170

0lhi:
@ds-sloth Are you gonna do #190 before merge?

LoveBodhi:
848x480 (424x240, like Hello's fangames) will be added?

0lhi:
Noticed a few issues:

* Battery is shown even when ```battery-status = "off"``` is set.

Battery

* Episode Title is not displayed anymore:

```
show-episode-title-small = 2
show-episode-title = 1
```

New
Old

ds-sloth:
That is not the system battery status, but the controller battery status. They're shown in different locations, controller next to the controls preview, and system in the upper right. (The reported controller battery status is the same as system battery status for keyboard and touchscreen.) I don't think controller battery status is on by default but I can check. Can you verify that disabling battery status in the controller profile settings resolves this?

Ah, yes. I'll backport the episode title code soon.

0lhi:
> I don't think controller battery status is on by default but I can check. Can you verify that disabling battery status in the controller profile settings resolves this?

Ah yeah, I missed that. It is on by default but disabling in the profile fixes it 👍

LoveBodhi:
But also:

- 640x360
- 848x480

ds-sloth:
Hmm. Those can be set using the custom resolution setting in the ini. I tried to include the resolutions of the systems that the source games were on (3DS is included because it's a target platform of TheXTech). If a sizable user base also wants Hello resolution (768 x 432) I can also add that as a default resolution.

Why are 640x360 and 848x480 interesting to you?

LoveBodhi:
These are modern 16:9 ratio!

ds-sloth:
Ah, I see. I wonder... one option would be to have all the preset settings, but have a separate setting called "auto screen width" or something like that. This would turn any vertical resolution into a dynamic resolution like the "600p dynamic" setting. Then, we would not have the 600p dynamic, only the fully dynamic. But any fixed resolution could become horizontally dynamic.

ds-sloth:
Thanks for the suggestions earlier, @0lhi!

> Battery is shown even when `battery-status = "off"` is set.

New profiles now read the main `config.ini` setting for joysticks to determine the default value. (That setting is off by default.)

> Episode Title is not displayed anymore:

I backported this feature now, but I use the same setting for the small-screen and big-screen cases. (Even though the location is still different -- screen top in big-screen, screen bottom in small-screen. Maybe we think about whether that's desirable.) I decided that, generally, users don't switch screen sizes frequently enough that it merits making one setting into either a complicated setting or a pair of settings. Also, the format has changed. It'll read the old numerical settings fine, but it'll also read "on"/"off"/"transparent" as strings.

Both of these have been updated both at the main branch and here.

0lhi:
> I decided that, generally, users don't switch screen sizes frequently enough that it merits making one setting into either a complicated setting or a pair of settings.

I mean, there definitely will be frequent switches between HD because it's cool and Vanilla cause some levels will be broken by larger resolutions.

I get wanting to avoid complications in the settings, but having one title setting for all sizes is flawed.

> (Even though the location is still different -- screen top in big-screen, screen bottom in small-screen. Maybe we think about whether that's desirable.)

Imo the small-screen titles are not desirable at all. I can't think of a scenario in which anyone would want to use them. For large-screens I requested them to have a convenient way for Streamers to let Viewers know the episode played. But on small-screens, they get in way of the view, which is not worth it.

If you want to keep things simple, I suggest one of the following approaches:

1. Remove the small-screen title altogether (and rename the setting to show-title-on-large-screen).
2. Make small-screen automatically transparent and large-screen automatically opaque. Add a Hotkey to hide/show the title.

---

On the Integer Setting with the Window-size 1280x720, after typing *smbxview*, the Window looks like this:

Scr_2022-05-28_15-00-09


While with Mode 2/3, it looks like this:

Scr_2022-05-28_14-58-34

Could you make the Integer setting behave like Mode 2/3? With the Framerate, Controller Input etc. being at the outer parts of the window, and Backdrop.png visible?

ds-sloth:
No, that isn't possible from a programming standpoint. If we want to have all those things onscreen, then we need to have a canvas that fits them. Your use case is specifically where users should set a compat flag for their level and get the view you showed in the lower screenshot.

So it's our job to make it easy to set this compat flag -- and in level settings are coming.

But also, the goal is to have as few levels as possible broken by multires by default. If you've seen anything that breaks gameplay, please report it here.

0lhi:
> and people were equally likely to want to not have it at all on smaller screens

Then, just remove it altogether? Or can you think of a scenario where people would want it?

I... really can't.

Scr_2022-05-28_16-45-22

ChristianSilvermoon:
Having it be transparent on small screens would be fine, I think. I wouldn't personally prefer it tho.

ChristianSilvermoon:
> I would keep the SDL_SetWindowMinimumSize() once to set the smallest size of the window (something ultra-small), just to avoid the user applying the impossibly small size. But for just window resizing and initial size it's fine to use the SDL_SetWindowSize()

image
You mean I couldn't play the world's smallest game of TheXTech anymore?
~~actually, this is still strangely playable...~~

Wohlstand:
>  You mean I couldn't play the world's smallest game of TheXTech anymore?

Try to play the game with 10x10 resolution :smile_cat:

ChristianSilvermoon:
:sunglasses:

https://user-images.githubusercontent.com/30162808/170840875-197072ca-7186-4525-9a44-414a6cbb3e23.mp4

~~In all honest; yeah this was really hard XD~~
~~Did manage to dodge some fireballs, tho!~~

ds-sloth:
Alright @0lhi, I've added the six settings for episode title. I still don't think most users are changing their resolution as frequently as you are, and I hope that users will use the compat setting to fix the level gameplay area instead of using the game's resolution itself. But it's not hard to have the options, and we can always remove them later if we ever do some telemetry to profile which options users actually choose.

0lhi:
> I've added the six settings for episode title.

Nice, they all work well. Once the Outline for Gameplay Window is available, I'll get into testing episodes on-stream.

0lhi:
At Mode 2 and 3, the Game-Window size can't be bigger than 800x600, but it can be smaller. This creates two problems:

1. Inventory Items can be accessed quicker.
2. Projectiles by Enemies get fired later.

Both create an unfair advantage. If possible, it would be good if the compat setting had 3 options:

```
limit-to-800x600 = no (default), yes, force
```

Force would give a result like this:

Mode2-Small

For 0.5 GBA it would be less than ideal, but necessary to create an even playfield.

If that's not possible, resolutions below 800x600 need to be locked out from using Mode 2/3. The simplest way to do this would be to notify the player when a resolution is to small, and automatically switch to Mode 1 upon starting an Episode.

ds-sloth:
Unless something is very different than I expect on Wii, every platform that we fully support should be *able* to do low-quality scaling like you demonstrated on the right. It'd look bad, but that's worth it if the player really wants Mode 2/3.

However, knowing the screen width and height at compile time does give a performance boost. The question is whether that performance boost is significant enough that we'd want to do fixed-res by default and offer a separate "speed-running build" that can force SMBX resolution. I agree that any fixed-resolution build with a <800x600 resolution should refuse to enter Mode 2/3, or should show "Mode 2R"/"Mode 3R" to signify that they aren't at the correct resolution.

I'll (1) force the resolution to be >=800x600 during Mode 2/3 for non-fixed res builds, and (2) figure out whether it's worth it to make a fixed res build (with the "Mode 3R" watermark) when I'm updating the 3DS port.

mikhailter:
Got this on my Nvidia Shield TV (Android 11):

version 1.3.5.3:

`Debug: Found 2 touch devices, screen size: 1920 x 1080`

Current pull request:

`Debug: Found 2 touch devices, screen size: 240 x 160`

I can't operate this tiny screen to change resolution to something reasonable :D Is the option to hard set resolution during first start?

ds-sloth:
Thanks for this very helpful report! I assume that something is wrong with the screen size detection on Android. Previously the app didn't support tiny resolutions at all and set the minimum resolution to 800x600, which must have hidden some of these issues.

Are you having the problem when using your Shield as a tablet, or when it's plugged in to a second screen?

mikhailter:
@ds-sloth 
> Are you having the problem when using your Shield as a tablet, or when it's plugged in to a second screen?

This is a TV box https://en.wikipedia.org/wiki/Nvidia_Shield_TV and it can be used with external display only.

ds-sloth:
I see, I'll try my best to reproduce this and fix it, but I don't have any comparable hardware myself. I think the only window initialization difference is the minimum window size, but there might also be some other flags. I'll look at the code more closely when I have time. Sorry that you aren't able to fully enjoy the game on your hardware yet!

0lhi:
Forced 800x600 on Mode 2/3 doesn't work in 2-Player:

Scr_2022-06-18_12-03-25

ChristianSilvermoon:
> Forced 800x600 on Mode 2/3 doesn't work in 2-Player:
Interesting split though x3

LoveBodhi:
Will add support for less than 800x600 section for new levels?

ds-sloth:
Hi @mikhailter , can you check if the change I just pushed ("Removed incorrect initial window size for Android") fixes the issue on your hardware? Sorry about the delay in addressing this issue. I don't have a Shield TV so it is hard for me to test out the changes myself.

ds-sloth:
I believe I've addressed all the major concerns raised since last time I've been working on the project. I haven't yet added code to force >800x600 during Mode 2/3 speedrun.

What other changes should be addressed before this can be merged? Of course some additional cases (esp. NPCs that should or should not spawn when onscreen) will probably arise once many people begin testing the new feature.

0lhi:
I think the last thing needed before merge is either smart world map screen crop, or alternatively limiting map size to 668x404, maybe with x2 Scale.

After Merge, there should be a separate issue with the dynamic stuff left to do (so I can report NPC Errors there).

LoveBodhi:
Scr_2022-07-03_07-01-15

The only problem is, It will be unusable when set to custom resolution (e.g. Hello Resolution at above) in the new editor.

LoveBodhi:
I will give a name of 768x432 for "Hello".

mikhailter:
> Hi @mikhailter , can you check if the change I just pushed ("Removed incorrect initial window size for Android") fixes the issue on your hardware? Sorry about the delay in addressing this issue. I don't have a Shield TV so it is hard for me to test out the changes myself.

Hi, i've checked latest build on my setup and it's have better resolution (it's possible to operate the menu), but still not running on fullscreen. I still need to send in to background and return back during loading to have it run on fullscreen without other manipulations.

p.s. please check the screen:
The X-Tech_20220704_234841

ds-sloth:
> Would it be possible to have an algorithm that fills these?

Sadly, I think not. It's relatively easy to do a reachability analysis on the world map because gameplay on the map is very constrained -- this is not the case in-game. This would be a place where somebody should either update the assets, or set a compat flag with a maximum supported resolution.

0lhi:
I see, yeah. I have another idea in mind that I'd like to bring up to you in private. Could you join Matrix?

ds-sloth:
@mikhailter I'll look more at this soon. Sorry that this bug has been affecting your gameplay for so long. I have an Android phone and will see how well I can reproduce the bug there. I'm interestingly getting some similar-looking issues when I run the game (either multires or the main branch) in SDL fullscreen on Wayland on Linux, but I'm not entirely certain why.

ds-sloth:
@0lhi I think the last thing on our list was the way to handle the world map cropping. How does something like this strike you?
Scr_2022-07-17_18-22-39

0lhi:
> @0lhi I think the last thing on our list was the way to handle the world map cropping. How does something like this strike you?

Gonna paste what I responded in Chat :P
```
:o 
That is super creative!
I think, the non-transparent grey could use more variety
What if it's white instead of grey?
```

anengineer1:
Yo! 0lhi contacted me to test this pull on wayland and here's the result
image

ds-sloth:
Thanks for the note, @anengineer1 ! Unfortunately SDL2's poor handling of fullscreen with Wayland is an issue I've been dealing with for a while, and it also affects the main branch. I'm glad to hear that it works fine for you in windowed mode. Would you mind putting up an issue report regarding this problem in general so we can discuss it separately?

anengineer1:
> Thanks for the note, @anengineer1 ! Unfortunately SDL2's poor handling of fullscreen with Wayland is an issue I've been dealing with for a while, and it also affects the main branch. I'm glad to hear that it works fine for you in windowed mode. Would you mind putting up an issue report regarding this problem in general so we can discuss it separately?

On your fork or Wohlstand's?

Wohlstand:
Also depends on which SDL2 to use, in-system, vendored, or custom-installed latest

Wohlstand:
A little note is that sometimes I contribute to SDL2 when I find some bugs by myself and when I find the solution by myself.

ds-sloth:
> On your fork or Wohlstand's?

This is the official repository for TheXTech -- I believe my fork only applies to the 3DS version.

ds-sloth:
@mikhailter I can't reproduce this issue on my Android phone. Can you confirm whether a *current* main build has the same issue?

mikhailter:
> @mikhailter I can't reproduce this issue on my Android phone. Can you confirm whether a _current_ main build has the same issue?

The issue present on the latest pull request build but latest official build v1.3.5.3 doesn't have this issue. I can send you a screen recording, but I need to contact you privately for this.

ds-sloth:
> The issue present on the latest pull request build but latest official build v1.3.5.3 doesn't have this issue. I can send you a screen recording, but I need to contact you privately for this.

Oh, I totally believe you! Can you check this build too please? https://github.com/Wohlstand/TheXTech/actions/runs/2715705390

Can you also remind me what happens with the multires build if you navigate to the menu, set a specific resolution like HD (rather than dynamic), and then restart the application?

LoveBodhi:
The cheat "helloview" should be added, too.

ds-sloth:
To my knowledge, these last commits solve the final remaining issues before merge. I'll be happy to fix any bugs and discuss any code review. Thanks to everyone for the very useful feedback!

0lhi:
To-Do-List for me:

- [ ] Confirm that Speedrun resizes work as intended in each Mode.
- [ ] Confirm that Piranha Plant Issue was resolved.
- [ ] Test default settings for inactive NPC handling.
- [ ] Test a few .ini configs.

ChristianSilvermoon:
For lower Resolutions, I think it's important to be able to pan the camera.

Many platformers allow panning up/down while holding up or down.

Right now the current behavior is to always center the player, but this leads to a lot of jumping to platforms you can't necessarily see ahead of time.

I think it would be good if holding still and up or down (for a period of 500+ ms) would change how the camera (whilst playing `sound-13`) follows the player, centering them at the top or bottom, like so:

image

Someone also suggested added a new a dedicated camera control button to enable panning up/down/left/right
While I do think this would be a good thing, I think it's important that the feature not be locked behind any additional inputs, so that devices with fewer buttons can still take advantage of camera panning (even if only vertical camera panning)

Some option to keep the player more to the left or the right of the screen depending on which direction they're facing and moving may also be ideal. For example, if they start moving left, keep them 3/4s of the way to the right side of the screen instead of centered.

So a sort of **Dynamic Horizontal Camera** option, that will gradually change how the camera horizontally centers the player based on movement and direction, and a **Manual Vertical Camera** option that changes how the camera vertically centers on the player based on holding up, or down while not moving for a brief period of time.

ds-sloth:
@ChristianSilvermoon Great writeup! I can't promise I'll get to this before merge because it's a relatively isolated case, but I'll think carefully about this and post some mockups in the chat. Maybe you could file another issue for this (just copy-paste it) so we can discuss it in a dedicated location and in case we don't address it before merge?

ChristianSilvermoon:
> @ChristianSilvermoon Great writeup! I can't promise I'll get to this before merge because it's a relatively isolated case, but I'll think carefully about this and post some mockups in the chat. Maybe you could file another issue for this (just copy-paste it) so we can discuss it in a dedicated location and in case we don't address it before merge?

<p>Animated GIF of Starlight Glimmer from My little Pony: Friendship is Magic nodding</p>
It is done #362 :3

Alucard648:
Have fun dodging enemy projectiles fired from off screen at lower resolutions.

0lhi:
> Have fun dodging enemy projectiles fired from off screen at lower resolutions.

When I tested the Fire Plant wouldn't shoot at small resolutions even though the same spot would trigger it at large resolutions. You should get the same result when testing The Invasion 2's first level.

(It's different here than in SMBX2, where the lower resolution is just an optical illusion.)

0lhi:
> To-Do-List for me:
> 
>     * [x]  Confirm that Speedrun resizes work as intended in each Mode.
>     * [x]  Confirm that Piranha Plant Issue was resolved.
>     * [x]  Test default settings for inactive NPC handling.
>     * [ ]  Test a few .ini configs.

So the first three are good.

Regarding the ```.ini```, I really like this :P

https://user-images.githubusercontent.com/74732674/181827448-b13216b6-7791-425f-8d06-212a65e96cb2.mp4

When I tried to add some random values for ```Backdrop_Border.png```, this happened:

Border

ds-sloth:
Thanks for the report @0lhi ! There was a bug where it couldn't handle an external size of zero, and also a possible creator mistake where the creator requests a total horizontal dimension (le + li + ri + re) greater than the image's width, and same on vertical axis. I also added a load-time warning for that situation.

0lhi:
Coolio, didn't crash, and setting everything to 2 with the cropped Border leads to the desired result!

I approve for merge 🙂👍

ChristianSilvermoon:
Oh  yeah, what about the title screen oddities?
Scr_2022-07-29_14-18-36

This still occurs with 800x600 title screen levels

ds-sloth:
The title screen issue should be fixed!

src/change_res.cpp
```diff
@@ -56,3 +68,193 @@ void ChangeRes(int, int, int, int)
 //{
 
 //}
+
+void UpdateInternalRes()
+{
+    int req_w = g_config.InternalW;
+    int req_h = g_config.InternalH;
+    if((!g_compatibility.free_level_res && !LevelSelect && !GameMenu)
```
Wohlstand, line 76: I would add the space between the definitions block and if condition.

src/change_res.cpp
```diff
@@ -56,3 +68,193 @@ void ChangeRes(int, int, int, int)
...
+        }
+        else if(g_videoSettings.scaleMode == SCALE_FIXED_05X)
+        {
+            int_h *= 2;
+        }
+        else if(g_videoSettings.scaleMode == SCALE_FIXED_2X)
+        {
+            int_h /= 2;
+        }
+        else if(g_videoSettings.scaleMode == SCALE_DYNAMIC_INTEGER)
```
Wohlstand, line 115: Braces can be removed here, for the single line they make not so much sense.

src/change_res.cpp
```diff
@@ -56,3 +68,193 @@ void ChangeRes(int, int, int, int)
...
+    }
+
+#ifndef THEXTECH_FIXED_RES
+    if(req_w == 0 || req_h == 0)
+    {
+        int int_w, int_h;
+
+        XWindow::getWindowSize(&int_w, &int_h);
+
+        int orig_int_h = int_h;
```
Wohlstand, line 100:
```cpp
int int_w, int_h, orig_int_h;

XWindow::getWindowSize(&int_w, &int_h);
orig_int_h = int_h;
```
would be cleaner here.

src/change_res.cpp
```diff
@@ -56,3 +68,193 @@ void ChangeRes(int, int, int, int)
...
+        {
+            int_h /= 2;
+        }
+        else if(g_videoSettings.scaleMode == SCALE_DYNAMIC_INTEGER)
+        {
+            if(int_h >= 600)
+            {
+                // constrains height to be in the 600-720p range
+                int scale_factor = int_h / 600;
+                int_h /= scale_factor;
```
Wohlstand, line 121: Why not just
```cpp
int_h /= (int_h / 600);
```
?

src/change_res.cpp
```diff
@@ -56,3 +68,193 @@ void ChangeRes(int, int, int, int)
...
+        }
+        else if(g_videoSettings.scaleMode == SCALE_FIXED_1X)
+        {
+            int_w = int_w;
+        }
+        else if(g_videoSettings.scaleMode == SCALE_FIXED_2X)
+        {
+            int_w = int_w / 2;
+        }
+        else if(g_videoSettings.scaleMode == SCALE_DYNAMIC_INTEGER)
```
Wohlstand, line 146: Also redundant braces for single-line code blocks.

src/change_res.cpp
ds-sloth: They are there because the final clause is multiline and I didn't want to have a bunch of unbraced followed by a braced multiline.

src/change_res.cpp
ds-sloth: This makes it clearer what the purpose of the `(int_h / 600)` term is, and the fact that truncation division was requested.

src/change_res.cpp
ds-sloth: Same case as above, the last clause is multiline and I want them to be balanced.

src/game_main.cpp
```diff
@@ -162,6 +163,28 @@ static std::string findIntroLevel()
...
+    for(int A = 1; A <= numBackground; A++)
+    {
+        if(CheckCollision(Background[A].Location, tempLocation))
+            return;
+    }
+
+    // expand level height to a maximum of 2160px
+    if(level[Player[1].Section].Y > level[Player[1].Section].Height - 2160)
+        level[Player[1].Section].Y = level[Player[1].Section].Height - 2160;
+}
```
Wohlstand, line 187: Please avoid so many complex deep expressions repeated, please make the reference:
```cpp
auto &section = level[Player[1].Section];
```
and then, re-use the `section` reference. This thing will be easier to maintain the logic.

src/gfx.cpp
```diff
@@ -123,7 +125,61 @@ bool GFX_t::load()
...
+
+    if(WorldMapFrame_Border.inited)
+    {
+        // warn if invalid
+        const FrameBorderInfo& i = g_worldMapFrameBorderInfo;
+        if(i.le + i.li + i.ri + i.re > WorldMapFrame_Border.w)
+            pLogWarning("Invalid border: total internal/external width is %d but WorldMapFrame_Border.png is only %dpx wide.", i.le + i.li + i.ri + i.re, WorldMapFrame_Border.w);
+        if(i.te + i.ti + i.bi + i.be > WorldMapFrame_Border.h)
+            pLogWarning("Invalid border: total internal/external height is %d but WorldMapFrame_Border.png is only %dpx wide.", i.te + i.ti + i.bi + i.be, WorldMapFrame_Border.h);
+    }
```
Wohlstand, line 166: Probably reasonable to unload this image as it doesn't meet the required conditions to just don't use it at all. After this warning, the image will still remain loaded and used. Or why do you keep it remaining?

src/graphics.cpp
```diff
@@ -105,9 +249,35 @@ void GetvScreen(const int A)
...
                 vScreen[A].TempY += 1;
         }
-        if(p.Mount == 2)
-            pLoc.Height = 128;
     }
+
+    // keep vScreen boundary even on 3DS
+#ifdef __3DS__
+    vScreenX[A] -= std::fmod(vScreenX[A], 2.);
+    vScreenY[A] -= std::fmod(vScreenY[A], 2.);
```
Wohlstand, line 261: Please don't use floating point numbers without zeroes. I mean, using `2.0` or `2.0f` or in the worst case `2.f` is fine, but `2.` is not so good.

src/graphics.cpp
```diff
@@ -105,9 +249,35 @@ void GetvScreen(const int A)
...
+
+    *left = -pLoc.X + (800 * 0.5) - pLoc.Width / 2.0;
+    *top = -pLoc.Y + (600 * 0.5) - vScreenYOffset - pHeight;
+    *left -= vScreen[A].tempX;
+    *top -= vScreen[A].TempY;
+    if(-(*left) < level[p.Section].X)
+        *left = -level[p.Section].X;
+    else if(-(*left) + 800 > level[p.Section].Width)
+        *left = -(level[p.Section].Width - 800);
+    if(-(*top) < level[p.Section].Y)
```
Wohlstand, line 281: Please add blank line between two different condition branches, they shouldn't be mixed.

src/graphics.cpp
```diff
@@ -153,17 +322,28 @@ void GetvScreenAverage()
             return;
         }
     }
-    vScreenX[1] = (vScreenX[1] / B) + (ScreenW * 0.5);
-    vScreenY[1] = (vScreenY[1] / B) + (ScreenH * 0.5) - vScreenYOffset;
-
-    if(-vScreenX[A] < level[Player[1].Section].X)
+    vScreenX[1] = (vScreenX[1] / B) + (vScreen[1].Width * 0.5);
```
Wohlstand, line 329: Blank line needs between brace and following code :eyes:

src/graphics.cpp
```diff
@@ -176,7 +356,61 @@ void GetvScreenAverage()
...
             vScreenX[1] = OldX - 10;
+
+        // on menu, bottom of screen always tracks bottom of level
+        vScreenY[A] = -(level[Player[1].Section].Height - vScreen[A].Height);
     }
+
+    // keep vScreen boundary even on 3DS
+#ifdef __3DS__
+    vScreenX[A] -= std::fmod(vScreenX[A], 2.);
+    vScreenY[A] -= std::fmod(vScreenY[A], 2.);
```
Wohlstand, line 371: ~~2.~~  -> `2.0`

src/graphics.cpp
```diff
@@ -176,7 +356,61 @@ void GetvScreenAverage()
...
+    {
+        if(GameMenu)
+        {
+            *left = -level[0].X;
+            B = 1;
+        }
+        else
+            return;
+    }
+    *left = (*left / B) + (800 / 2);
```
Wohlstand, line 407: Closing brace, and folowing code should have a blank line

src/graphics.cpp
```diff
@@ -176,7 +356,61 @@ void GetvScreenAverage()
...
+    *top = (*top / B) + (600 / 2) - vScreenYOffset;
+
+    if(-(*left) < level[Player[1].Section].X)
+        *left = -level[Player[1].Section].X;
+    else if(-(*left) + 800 > level[Player[1].Section].Width)
+        *left = -(level[Player[1].Section].Width - 800);
+    if(-(*top) < level[Player[1].Section].Y)
```
Wohlstand, line 414: Missing blank line between different condition branches.

src/graphics.cpp
```diff
@@ -207,6 +441,45 @@ void GetvScreenAverage2()
 
     vScreenX[1] = (vScreenX[1] / B) + (ScreenW * 0.5);
     vScreenY[1] = (vScreenY[1] / B) + (ScreenH * 0.5) - vScreenYOffset;
+
+#ifdef __3DS__
+    vScreenX[1] -= std::fmod(vScreenX[1], 2.);
+    vScreenY[1] -= std::fmod(vScreenY[1], 2.);
```
Wohlstand, line 451: `2.` -> `2.0`

src/graphics.cpp
```diff
@@ -207,6 +441,45 @@ void GetvScreenAverage2()
...
+    for(int A = 1; A <= numPlayers; A++)
+    {
+        if(!Player[A].Dead)
+        {
+            l += -Player[A].Location.X - Player[A].Location.Width / 2.0;
+            if(Player[A].Mount == 2)
+                t += -Player[A].Location.Y;
+            else
+                t += -Player[A].Location.Y - Player[A].Location.Height;
+            B += 1;
```
Wohlstand, line 472: Maybe `++B`?

src/graphics/gfx_background.cpp
```diff
@@ -22,963 +22,667 @@
...
+    {
+        XRender::renderRect(0, 0, vScreen[Z].Width, vScreen[Z].Height,
+            GFXBackground2[A].ColorLower.r, GFXBackground2[A].ColorLower.g, GFXBackground2[A].ColorLower.b);
+    }
+    else
+    {
+        XRender::renderRect(0, 0, vScreen[Z].Width, vScreen[Z].Height,
+            GFXBackground2[A].ColorUpper.r, GFXBackground2[A].ColorUpper.g, GFXBackground2[A].ColorUpper.b);
+    }
+}
```
Wohlstand, line 40: You can simplify the whole thing if you will use the reference:
```cpp
const auto &color = lower ? GFXBackground2[A].ColorLower : GFXBackground2[A].ColorUpper;
XRender::renderRect(0, 0, vScreen[Z].Width, vScreen[Z].Height, color.r, color.g, color.b);
```

src/graphics/gfx_hud.cpp
```diff
@@ -294,134 +341,145 @@ void DrawInterface(int Z, int numScreens)
         {
             if(Player[1].Character == 3 || Player[1].Character == 4 || Player[1].Character == 5)
             {
-//                BitBlt myBackBuffer, vScreen[Z].Width / 2.0 - GFX.Heart[1].w / 2 + C - 32, 16, GFX.Heart[1].w, GFX.Heart[1].h, GFX::HeartMask(1).hdc, 0, 0, vbSrcAnd;
-//                BitBlt myBackBuffer, vScreen[Z].Width / 2.0 - GFX.Heart[1].w / 2 + C, 16, GFX.Heart[1].w, GFX.Heart[1].h, GFX::HeartMask(1).hdc, 0, 0, vbSrcAnd;
-//                BitBlt myBackBuffer, vScreen[Z].Width / 2.0 - GFX.Heart[1].w / 2 + C + 32, 16, GFX.Heart[1].w, GFX.Heart[1].h, GFX::HeartMask(1).hdc, 0, 0, vbSrcAnd;
+//                BitBlt myBackBuffer, CenterX - GFX.Heart[1].w / 2 + C - 32, 16, GFX.Heart[1].w, GFX.Heart[1].h, GFX::HeartMask(1).hdc, 0, 0, vbSrcAnd;
+//                BitBlt myBackBuffer, CenterX - GFX.Heart[1].w / 2 + C, 16, GFX.Heart[1].w, GFX.Heart[1].h, GFX::HeartMask(1).hdc, 0, 0, vbSrcAnd;
+//                BitBlt myBackBuffer, CenterX - GFX.Heart[1].w / 2 + C + 32, 16, GFX.Heart[1].w, GFX.Heart[1].h, GFX::HeartMask(1).hdc, 0, 0, vbSrcAnd;
```
Wohlstand, line 346: Did editing comment code? :)

Doesn't so matter, this code block was left a while ago as a reference for some and forgotten...

src/graphics/gfx_screen.cpp
```diff
@@ -213,19 +252,31 @@ void DynamicScreen()
...
                 }
-                vScreen[2].Visible = true;
+
+                if(DScreenType != 3)
+                    PlaySound(SFX_Camera);
                 DScreenType = 3;
+                vScreen[2].Visible = true;
             }
-            else if(level[Player[1].Section].Height - level[Player[1].Section].Y > ScreenH && ((!vScreen[2].Visible && Player[2].Location.Y + vScreenY[1] >= ScreenH * 0.75 - vScreenYOffset - Player[2].Location.Height) || (vScreen[2].Visible && Player[2].Location.Y + vScreenY[1] >= ScreenH * 0.75 - vScreenYOffset - Player[2].Location.Height)) && (Player[1].Location.Y < level[Player[1].Section].Height - ScreenH * 0.75 - vScreenYOffset - Player[1].Location.Height))
+            else if(level[Player[1].Section].Height - level[Player[1].Section].Y > FieldH && ((!vScreen[2].Visible && Player[2].Location.Y + vScreenY[1] >= FieldH * 0.75 - vScreenYOffset - Player[2].Location.Height) || (vScreen[2].Visible && Player[2].Location.Y + vScreenY[1] >= FieldH * 0.75 - vScreenYOffset - Player[2].Location.Height)) && (Player[1].Location.Y < level[Player[1].Section].Height - FieldH * 0.75 - vScreenYOffset - Player[1].Location.Height))
```
Wohlstand, line 279: So messy conditions suggested to be split into multiple lines, making less pain for eyes :eyes:

src/graphics/gfx_special_frames.cpp
```diff
@@ -116,3 +116,131 @@ void SpecialFrames()
...
+    {
+        BackgroundFrame[26]++;
+        if(BackgroundFrame[26] >= 8)
+            BackgroundFrame[26] = 0;
+        BackgroundFrameCount[26] = 0;
+    }
+    BackgroundFrameCount[18]++;
+    if(BackgroundFrameCount[18] >= 12)
+    {
+        BackgroundFrame[18]++;
```
Wohlstand, line 132: Please add blank lines between conditions and lonely code lines :eyes:

Wohlstand (commented):
I reviewed almost everything quickly, I noted some code-style and optimisation cases, in the next evening I'll get a shot to review another part of the code.

src/core/base/render_base.cpp
```diff
@@ -236,15 +236,15 @@ StdPicture AbstractRender_t::LoadPicture(const std::string &path,
...
+    target.ColorUpper.b = upperColor.rgbBlue / 255.0f;
 
     RGBQUAD lowerColor;
     FreeImage_GetPixelColor(sourceImage, 0, static_cast<unsigned int>(h - 1), &lowerColor);
-    target.ColorLower.r = lowerColor.rgbRed;
-    target.ColorLower.b = lowerColor.rgbBlue;
-    target.ColorLower.g = lowerColor.rgbGreen;
+    target.ColorLower.r = lowerColor.rgbRed / 255.0f;
+    target.ColorLower.b = lowerColor.rgbBlue / 255.0f;
+    target.ColorLower.g = lowerColor.rgbGreen / 255.0f;
```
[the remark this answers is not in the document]
ds-sloth: As far as I could tell, they were `float` fields, invalidly being treated as if they were `uint8_t`s. I had to change this to get the background fill-in to work correctly. I'd personally prefer that we make a class that treats colors as (MSByte)rgba(LSByte) `uint32_t`s, with `constexpr` constructors that take `float`s or `uint8_t`s, but this can be a future change.

src/change_res.cpp
```diff
@@ -56,3 +68,193 @@ void ChangeRes(int, int, int, int)
...
+
+    XRender::updateViewport();
+
+    // recenter the game menu graphics
+    if(GameMenu)
+    {
+        SetupScreens();
+        GameMenu = false;
+        GetvScreenAverage();
+        GameMenu = true;
```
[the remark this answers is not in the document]
ds-sloth: I didn't want to add the argument since it is almost always the correct behavior to use `GameMenu` and looking it up as a global before the function call and passing it as an argument is more expensive than directly looking up the global in the function. But we can also change it to an argument if you think this is very bad style, or I can add a comment indicating why this was done this way.

src/std_picture.h
```diff
@@ -33,11 +33,14 @@
...
+// 32-bit integer for efficient parameter passing;
+// see the wip render queue branch
 struct PGEColor
 {
-    uint8_t r = 0;
-    uint8_t g = 0;
-    uint8_t b = 0;
+    float r = 0;
+    float g = 0;
+    float b = 0;
```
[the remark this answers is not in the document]
ds-sloth: Hmm, let's talk more about this soon. I think it's possible that some incompatible code mixing took place when I initially rebased the feature on the current main branch, because I remember my initial code mixing `float`s and `uint8_t`s. After this merge, I can make a mini PR that converts all colors into `uint32_t`-backed classes as I discussed in the previous comment.

src/game_main.cpp
ds-sloth: Better still, I see now that some code mixing happened here. It's all meant to be `level[0]`, though they are the same on any valid `intro.lvl` files. I'll fix this. Thanks for the note!

src/gfx.cpp
ds-sloth: It would be reasonable to de-init the image here, yes. This check is also used internally in `gfx_frame.cpp`, so it wouldn't be used regardless, but I could save some CPU by de-initing. The main thing I don't remember is whether it is safe to de-init a custom GFX or not, since we would also want to be able to do that.

src/graphics.cpp
ds-sloth: I normally use postfix for primitives and prefix for literals, but I can use prefix here if you prefer. Agree that both are preferred to `+= 1`.

src/graphics/gfx_hud.cpp
ds-sloth: Ha, must have done a replacement :)

ds-sloth:
Thanks to @Wohlstand's hard work finalizing everything, 1.3.6 has been released!

As far as I can tell, there are two blocking issues before merge:
- Resolve some minor gameplay recording divergences I recently discovered
- Improve rendering for AutoCode HUDs that target 800x600. I think that >800x600 is already done, but <800x600 still needs work I think.

Does anyone else have any issues blocking merge?

Of course, we will continue to target performance and bugfix updates to the 1.3.6.1 branch even after merging multires.

Thanks to everyone who has contributed to the discussion of this PR!

src/main/menu_main.cpp
```diff
@@ -2040,14 +2040,14 @@ void mainMenuDraw()
     {
         int i = 0;
         SuperPrint(g_mainMenu.controlsTitle, 3, MenuX, MenuY + 30*i++);
-#ifndef __ANDROID__
+#ifndef RENDER_FULLSCREEN_ALWAYS
         if(resChanged)
-            SuperPrint("WINDOWED MODE", 3, MenuX, MenuY + 30*i++);
+            SuperPrint("WINDOWED MODE", 3, 300, 350 + (30 * i++));
         else
```
ds-sloth, line 2046: Best to roll these two lines back, they'll make a strange visual effect at non 800x600 resolutions

Wohlstand: Oh, I see now, didn't paid attention

Wohlstand: Fixed just now!

ds-sloth:
I just spent some time with Foret Bleue and A2XT and am now convinced that it's not possible to make a perfect way of rendering AutoCode content at non-800x600 resolutions.

It's not our job to make that perfect way as long as players understand that the content was designed for 800x600, which I know @0lhi has said is essential for a lot of other reasons as well.

This was the last pressing issue I was aware of with multires, so I think it is ready for merge whenever both I and @Wohlstand are happy with the code.

SkullGamer205:
👀

ds-sloth:
This PR has been marked as a draft because it will not be merged until after the upcoming 1.3.6.1 release.

ds-sloth:
Important note: this branch may become temporarily unstable as some important code reorganization is performed. Once the reorganization is over, the behavior will be the same as always. This work will simplify the internal logic and make it easier to implement multi-window or NetPlay support in the future.

Please let me know here or on Discord / Matrix if you notice any issues.

LoveBodhi:
Some new CN localizations:

- Resolution - 分辨率
- SNES - SFC

src/change_res.cpp
```diff
@@ -68,16 +74,176 @@ void ChangeRes(int, int, int, int)
...
+        else if(g_videoSettings.scaleMode == SCALE_FIXED_2X)
+        {
+            int_w = int_w / 2;
+        }
+        else if(g_videoSettings.scaleMode == SCALE_DYNAMIC_INTEGER)
+        {
+            int scale_factor = orig_int_h / int_h;
+            if(scale_factor == 0)
+            {
+                int_w = int_w;
```
Wohlstand, line 146: Why there is an assignment to self? You may want just put a comment here, otherwise compilers may drop a warning on this.

src/change_res.cpp
```diff
@@ -68,16 +74,176 @@ void ChangeRes(int, int, int, int)
...
+        {
+            int_w = int_w * 2;
+        }
+        else if(g_videoSettings.scaleMode == SCALE_FIXED_1X)
+        {
+            int_w = int_w;
+        }
+        else if(g_videoSettings.scaleMode == SCALE_FIXED_2X)
+        {
+            int_w = int_w / 2;
```
Wohlstand, line 139: `int_w /= 2;`

src/change_res.cpp
```diff
@@ -68,16 +74,176 @@ void ChangeRes(int, int, int, int)
...
+            int_h = 720;
+
+        // now, set width based on height and scaling mode
+        if(g_videoSettings.scaleMode == SCALE_FIXED_05X)
+        {
+            int_w = int_w * 2;
+        }
+        else if(g_videoSettings.scaleMode == SCALE_FIXED_1X)
+        {
+            int_w = int_w;
```
Wohlstand, line 135: Self-assignment is a source of compiler warnings, better just leave a comment here.

src/change_res.cpp
```diff
@@ -68,16 +74,176 @@ void ChangeRes(int, int, int, int)
...
+            int_h = 320;
+
+        // maximum height constraint
+        if(int_h > 720 && req_h <= 720)
+            int_h = 720;
+
+        // now, set width based on height and scaling mode
+        if(g_videoSettings.scaleMode == SCALE_FIXED_05X)
+        {
+            int_w = int_w * 2;
```
Wohlstand, line 131: `int_w *= 2`

src/blocks.cpp
```diff
@@ -2396,44 +2396,71 @@ void PSwitch(bool enabled)
...
+            {
+                if(ScreenType == 1)
+                    numScreens = 2;
+                if(ScreenType == 4)
+                    numScreens = 2;
+
+                if(ScreenType == 5)
+                {
+                    DynamicScreen(screen);
+                    if(vScreen[2].Visible)
```
Wohlstand, line 2434: Maybe use a ternary expression here?
```cpp
numScreens = vScreen[2].Visible ? 2 : 1;
```

Wohlstand (commented):
I made a small look, and I have some notes.

src/change_res.cpp
ds-sloth: Thanks for the comments here, just updated the style of the procedure.

src/blocks.cpp
ds-sloth: Good note, I wanted to keep the exact format since this is copied legacy logic. Do you still want me to update it?

Wohlstand: Ye, it would be better to update

ds-sloth: I updated the syle of this section (removed the `else` clause and combined the previous `if` statements)

Wohlstand:
As a little exception to our internal rule (i.e. review each other before merge), let's merge this without of my complete review and test it out soon?

Why? Because of freaking Dezember (I was ill, had a strong cold, and was almost unavailable: I staid in my bed and was asleep the rest of time), and because of that a lot of stuff around got a strong delay. That's not good. Totally.

Wohlstand:
Anyway, I did a partial review of the thing, and seems fine for me. I think, other parts I'll review in hot during experiments at the Main branch.

P.S. Do Squish merge.

Wohlstand (approved):
Approving this quickly, so, if anything I'll find, I'll fix that it being already merged to Main.

PR #342, opened 2022-06-18
Refactoring of the AppPath module

Wohlstand:
I made the refactoring of the AppPath module which is used to recognize global paths such as assets, user directory, and others.
So, now the module has two parts:
- Main part (app_path.cpp), the platform-independent code which uses the output of the second part. This module would ignore returned paths and use, for example, user-specified.
- Platform-dependent modules, there are a set of modules which return system default assets and user paths.

After the module gets merged, I have to slightly polish it. One thing is left - need to test the work of the thing on Vita (I made the code, but I am not sure will it work or not, because I don't have enough hardware to verify this).


Wohlstand:
@0lhi, could you test out this thing in action? (portable packages and DEB-based things)

Wohlstand:
Okay, it works, but needs future polishing, will do that later, but going to merge now.

PR #351, opened 2022-07-12
Fixed some NPC ID bugs I accidentally introduced

P-Star7:


src/npc_id.h
```diff
@@ -23,6 +23,359 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#ifndef ENUMNPCID_HHH
+#define ENUMNPCID_HHH
+
```
Wohlstand, line 378: 纳尼？ You made the file that contains two copies of self in the same file

Wohlstand (changes requested):

src/npc_id.h
LoveBodhi: Oh, shucks! Just a duplicate copy...!

Wohlstand:
This pull request is completely broken, please re-do it from the latest state.

PR #352, opened 2022-07-13
Adding JP Names

LoveBodhi:


LoveBodhi:
Maybe the conflict point is Notepad++?

LoveBodhi:
> You should reset your state to the current `master` branch state (pull all the latest changes) and re-apply your changes again. Then, force-push to your repository.

How do I reset state?

Wohlstand:
> > You should reset your state to the current `master` branch state (pull all the latest changes) and re-apply your changes again. Then, force-push to your repository.
> 
> How do I reset state?

```
git fetch mainstream
git reset --hard mainstream/master
git push --force origin master
```

ds-sloth:
It is good that you are tracking all of the Japanese names for the NPC types, but from a source-code management perspective it is better to have only a single name for each NPC type. This way we can easily search the entire repository for all of the code that affects an NPC.

Perhaps you could comment all of these lines out, so that we can remember them, but only use a single name for each NPC.

Wohlstand:
Also, I totally forgot to say: if you think these aliases are useful for scripts, they should be defined in the script binding code (once it will be built in the future), but shouldn't be defined here. So, for the script API these aliases can be useful, but they are complicating the maintenance of the main code.

Wohlstand:
So, my best tip: close this, and wait until I implement the initial lua API, and then we can add these aliases into the binding.

LoveBodhi:
Actually, These names are **EXCLUSIVE TO SCRIPTS ONLY**! NOT for other things.

PR #376, opened 2022-08-11
Block quadtree

ds-sloth:
The code here is not fully optimized and slightly decreases performance (~10%) on small and simple levels, BUT greatly increases performance on tall levels and levels with moving layers.

Once this is merged, we can freely experiment with the internal implementation and try mechanisms other than the Loose Quadtree. As long as there are not internal bugs in our later implementations, then there will be NO possible effect on gameplay from these experiments, only performance increase/decrease.


src/global_constants.h
```diff
@@ -69,6 +69,10 @@ const int maxStarsNum = 1000;
 
 const int maxLocalPlayers = 2;
 
+// was previously changed to 10000; now only used for certain NPC bounds checking.
+// returned to 8000.
+const int FLBlocks = 8000;
```
Wohlstand, line 74: I had to set 10000 not because of NPCs, this would lead the failure on bigger levels with 21+ sections (using the old algorithm). But, I guess, the new algorithm shouldn't limit the range at all.

ds-sloth: Right. Now it is used to ensure old (possibly buggy behavior) for certain cases where (if I remember right) NPC collision detection was disabled outside of the standard +/-8000 range.

Wohlstand: That condition I added by me to fix the crash that happens when some flying NPC gets boosted into the speed of lighting, and immediately causes the game to crash, and I added that check. It's not vanilla, It's mine. You can comment it freely.

src/graphics/gfx_update.cpp
```diff
@@ -1415,22 +1420,28 @@ void UpdateGraphics(bool skipRepaint)
...
+            -vScreenX[Z], -vScreenY[Z],
+            -vScreenX[Z] + vScreen[Z].Width, -vScreenY[Z] + vScreen[Z].Height,
+            SORTMODE_ID);
 
 
 //        For A = fBlock To lBlock 'Non-Sizable Blocks
-        For(A, fBlock, lBlock)
+        for(Block_t* block : screenBlocks)
         {
+            A = block - &Block[1] + 1;
```
Wohlstand, line 1444: I think, if there is no index-related formula, then use the block reference directly in the code below. I'll make this by myself later.

ds-sloth: Looks good. I could also switch the quadtree later to internally use and return `BlockRef_t`s, which will be more cache-friendly and be directly useful for both indexes and references.

Wohlstand: Good idea :+1:

src/script/luna/mememu.cpp
```diff
@@ -487,7 +487,7 @@ class SMBXMemoryEmulator
         insert(0x00B2C882, &MenuMode); // Current menu mode
 
         // insert(0x00B2C884, ???}; // Key Released!!!
-        insert(0x00B2C894, &BlocksSorted);
+        // insert(0x00B2C894, &BlocksSorted); // removed by block quadtree
```
Wohlstand, line 490: Better to give the fake field that will be always "true", just to not confuse the "ABI"

Wohlstand: Okay, I fixed it by myself :fox_face:

ds-sloth:
Thank you for the fix!

Wohlstand (approved):
After some checks and tweaks, it's fine to merge now.

Wohlstand:
If anything is needed, we can make that later.

PR #419, opened 2022-09-20
Working Nintendo Switch port via DevkitPro

Wohlstand:
This is an effort to make the game work on the Nintendo Switch natively.

However, there are some known issues:
- [ ] Touchscreen events don't react (is there an issue on the SDL2 side?)

Tested on Yuzu and earlier build on the hardware.


CMakeLists.txt
```diff
@@ -599,6 +613,14 @@ if(ANDROID)
     target_link_libraries(A2XT_Int INTERFACE log)
 endif()
 
+if(NINTENDO_SWITCH)
+    target_link_libraries(A2XT_Int INTERFACE nx)
+
+    # Set explicit flags for this project
+    target_compile_definitions(thextech PRIVATE
+        -D__SWITCH__ -DNO_SCREENSHOT -DFIXED_RES -DPRELOAD_LEVELS -DNO_INTPTROC)
```
ds-sloth, line 621: (1) Can it build without these? `-DNO_SCREENSHOT -DFIXED_RES`
(2) If files could possibly be modified when game is open, better to remove `-DPRELOAD_LEVELS` -- I think Switch filesystem is fast so it's fine to include this in the Menu Screen.

src/core/sdl/window_sdl.cpp
```diff
@@ -89,14 +89,26 @@ bool WindowSDL::initSDL(const CmdLineSetup_t &setup, uint32_t windowInitFlags)
 
     SDL_GL_ResetAttributes();
 
+#if defined(__SWITCH__) /* On Switch, expect the initial size 1280x720 */
+    const int initWindowW = 1920;
+    const int initWindowH = 1080;
```
ds-sloth, line 94: Should it be 1280x720? This requires less scaling. In docked mode, it might upscale to 1920x1080 then downscale to 1280x720. We can also review this later at multires time.

ds-sloth (approved):
I read everything, and it looks good. We can continue work on the SDL2 touchscreen issue after merging. I think this is good to merge to main and stable.

CMakeLists.txt
Wohlstand: Gonna make a shot...

Wohlstand: About screenshots, there is built-in screenshot functionality of Switch itself, including making short video recordings

src/core/sdl/window_sdl.cpp
Wohlstand: I did some tests, and it should keep the 1920x1080, but, when I set 1280x720, when it's upscaled, quality distorts

Wohlstand: I talked with Yuzu folks about this, and it's better to keep 1920x1080 for the best picture quality, and it's fine when it gets auto-scaled during portable mode.

ds-sloth: This is a pixel graphics game, so I'd prefer to keep the integer scaling as close to prefect as possible, but let's worry about that during Multires. Right now there is no integer scaling support at all.

PR #422, opened 2022-09-22
Performance 1

ds-sloth:
Hi Wohlstand,

Please review this PR. Features:

(1) uses much simpler and more efficient table data structure instead of the overly general loose quadtree
(2) includes a new table for BGOs
(3) allows layer tables to easily split from the main table when they start moving, and join back to the main table when they stop

Remaining tasks for the future:

(1) make tables for PEZ and warps
(2) align the tables to the standard SMBX64 32x32 grid that most levels follow
(3) improve NPC performance

I think that (1) and (2) can happen directly in the main branch, and I can open another PR for (3) in the future.


Wohlstand:
Gonna get a shot to fix the Windows build first, booting my VM...

src/player/player_update.cpp
```diff
@@ -3117,11 +3120,14 @@ void UpdatePlayer()
...
-                for(B = 1; B <= numBackground; B++)
+                for(int B : treeBackgroundQuery(Player[A].Location, SORTMODE_NONE))
                 {
+                    if(B > numBackground)
+                        continue;
+
                     if(BackgroundFence[Background[B].Type] && (!g_compatibility.fix_climb_invisible_fences || !Background[B].Hidden))
                     {
-                        if(CheckCollision(Player[A].Location, Background[B].Location))
+                        // if(CheckCollision(Player[A].Location, Background[B].Location))
```
Wohlstand, line 3130: Why did you remove the collision check with the vine BGOs? Are you absolutely sure player won't collide them outside the actual collision?

Wohlstand (commented):
All seems fine, but I made minor edits to the code style. But, I have one small question.

src/player/player_update.cpp
ds-sloth: I agree, that doesn't seem right. Let me look closer...

ds-sloth: Ah, right. Refer to the tempLocation collision below it, it was just a duplicated check that increased computational efficiency when collision was unlikely, but reduces computational efficiency when collision is likely.

Wohlstand (approved):
So, now seems fine to merge!

PR #447, opened 2022-12-11
Upgrade SDL Proxy and other parts

Wohlstand:
There is my work to polish the `wip-ports` branch. I desired to make the work at the second branch with the goal to simplify the review process.


ds-sloth:
Visually, all looks good. I skimmed everything. Nice style edits on my 3DS audio library. After we do the benchmarks for vanilla MixerX on 3DS, if it is slow enough that we need to continue to use my library, then I'll do a deeper rewrite to make it more object oriented with the different formats less hard coded. I'll try to make it work with the signatures from your AudioCodecs library where possible, I think it's very clean.

Do you want me to help fix the 3DS, Wii, and headless (command line POSIX / "null") builds?

Wohlstand:
I want to take a look, I had posted the stuff, and then try other parts too

PR #449, opened 2022-12-29
Updated font engine

### LINKED ISSUE #41: Backport the Moondust Engine's font engine to provide the multi-language and the unicode support
Since the foundation, the game uses the old vanilla font engine made by Redigit that can print ASCII only. This adds a lot of limitations and strongly prevents the translation of the game into other languages. Together with that, as a separate task, develop the episode translation system to allow non-damage translation of episodes (by making the separate translation files in addition to existing levels and world maps).

Wohlstand:
This is a new font engine taken from the Moondust Engine. It gives the ability to print Unicode text using the UTF8 charset.
Scr_2022-12-29_15-08-19

There is some stuff still left to test/polish.

Fixes #41


LoveBodhi:
Windows side isn't fixed yet...

Wohlstand:
I know, also on macOS, I'll work on this soon

src/fontman/raster_font.cpp
```diff
@@ -0,0 +1,490 @@
...
+            //Fake tabulation
+            offsetX += offsetX + offsetX % w;
+            continue;
+
+        case ' ':
+            offsetX += m_spaceWidth + m_interLetterSpace / 2;
+            continue;
+        }
+
+        RasChar rch = m_charMap[get_utf8_char(strIt)];
```
ds-sloth, line 440: I think this mallocs an entire RasChar for every new failed char. Is that right? Maybe better to do a find call instead.

src/fontman/raster_font.cpp
```diff
@@ -0,0 +1,490 @@
...
+    m_spaceWidth       = rf.m_spaceWidth;
+    m_newlineOffset    = rf.m_newlineOffset;
+    m_glyphOffsetX     = rf.m_glyphOffsetX;
+    m_glyphOffsetY     = rf.m_glyphOffsetY;
+    m_matrixWidth      = rf.m_matrixWidth;
+    m_matrixHeight     = rf.m_matrixHeight;
+    m_isReady          = rf.m_isReady;
+    m_ttfBorders       = rf.m_ttfBorders;
+    m_charMap          = rf.m_charMap;
+    m_texturesBank     = rf.m_texturesBank;
```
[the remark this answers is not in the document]
ds-sloth: Maybe we should enforce that the copy operator of StdPicture doesn’t reuse any internal data?

src/fontman/ttf_font.cpp
```diff
@@ -0,0 +1,484 @@
...
+    FT_GlyphSlot glyph  = cur_font->glyph;
+    FT_Bitmap &bitmap   = glyph->bitmap;
+    uint32_t width      = bitmap.width;
+    uint32_t height     = bitmap.rows;
+    uint32_t pitch      = width * 4;
+
+    if((width == 0) || (height == 0))
+        return dummyGlyph;
+
+    auto *image = new uint8_t[4 * width * height];
```
ds-sloth, line 418: No OOM handling

src/globals.h
```diff
@@ -147,6 +147,7 @@ extern bool ScrollRelease;
 extern bool TakeScreen;
 // EXTRA: Show any on-screen meta (HUD, debug prints, etc.)
 extern bool ShowOnScreenHUD;
+extern bool NewFontRender;
```
ds-sloth, line 150: Should document, including “EXTRA” tag

src/graphics/gfx_message.cpp
```diff
@@ -34,19 +39,36 @@ void DrawMessage(const std::string& SuperText)
     int lineStart = 0; // start of current line
     int lastWord = 0; // planned start of next line
     int numLines = 0; // n lines
-    int maxChars = (TextBoxW-24)/charWidth+1; // 27 by default
+    const int maxChars = ((TextBoxW - 24) / charWidth) + 1; // 27 by default
+
+    // Build map
+    std::vector<const char*> SuperTextMap;
```
ds-sloth, line 45: Better to do this when entering the message pause state I think instead of allocing and freeing every frame

src/graphics/gfx_message.cpp
```diff
@@ -1,7 +1,11 @@
-#include "globals.h"
+﻿#include "globals.h"
 #include "../graphics.h"
 #include "core/render.h"
 #include "../gfx.h"
+#include "fontman/font_manager_private.h"
+#include "fontman/font_manager.h"
+
+//#define USE_OLD_TEXT_PARSER
```
ds-sloth, line 8: Seems totally unused here?

src/graphics/gfx_print.cpp
```diff
@@ -44,11 +47,25 @@ int SuperTextPixLen(int SuperN, const char* SuperChars, int Font)
     default:
     case 1:
     case 4:
-        len = SuperN * 18;
+    {
+        //len = SuperN * 18;
+        for(int i = 0; i < SuperN; i++)
+        {
+            len += 18;
```
ds-sloth, line 54: Strictly, this should depend on the font override, yes?

src/graphics/gfx_print.cpp
```diff
@@ -75,25 +93,59 @@ int SuperTextPixLen(int SuperN, const char* SuperChars, int Font)
 
 void SuperPrintRightAlign(int SuperN, const char* SuperChars, int Font, float X, float Y, float r, float g, float b, float a)
 {
+    int dFont = FontManager::fontIdFromSmbxFont(Font);
+    if(NewFontRender && dFont >= 0)
+    {
+        const auto out = std::string(SuperChars).substr(0, SuperN);
```
ds-sloth, line 99: Probably better to make your functions accept sized buffers. There are cases in the config screen where I make views of ~40 strings and I’d rather they not be reallocated every frame.

ds-sloth: We could also consider upgrading everything to I use string_views but that’s a bit complex for me.

src/graphics/gfx_print.cpp
```diff
@@ -235,26 +300,63 @@ void SuperPrint(int SuperN, const char* SuperChars, int Font, float X, float Y,
 
 int SuperTextPixLen(const char* SuperChars, int Font)
 {
-    return SuperTextPixLen(SDL_strlen(SuperChars), SuperChars, Font);
+    int dFont = FontManager::fontIdFromSmbxFont(Font);
```
ds-sloth, line 303: For all of these, maybe better to check render mode before looking for font?

src/fontman/raster_font.cpp
```diff
@@ -0,0 +1,490 @@
...
+        std::u32string ucharX = std_to_utf32(charX);
+        char32_t ch = ucharX[0];
+        //qDebug()<<"=char=" << ch << "=id="<<charPosX.toInt()<<charPosY.toInt()<<"=";
+        RasChar rch;
+#ifndef DEBUG_BUILD
+        try
+        {
+#endif
+            rch.tx              =  loadedTexture;
+            rch.l               =  std::stof(charPosY.c_str()) / m_matrixWidth;
```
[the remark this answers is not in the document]
ds-sloth: Better yet, I’d rather just use a grid, and ONLY store grid index (row * n_cola + col) as a uint16_t. A bit less flexible, but MUCH easier to store. I want to be able to support as many scripts as possible even on our tiniest platforms.

src/main/screen_textentry.cpp
```diff
@@ -455,14 +490,14 @@ bool KeyboardMouseRender(bool mouse, bool render)
...
-    int kb_height = 5*key_size;
-    int kb_width = 12*key_size;
+    int kb_height = s_current_keymap_rows*key_size;
+    int kb_width = s_current_keymap_cols*key_size;
 
     int win_width = kb_width + 20;
     int win_height = kb_height + 20;
 
     int n_prompt_lines = ((s_Prompt.size() + 26) / 27);
-    int n_text_lines = ((Text.size() + 24) / 25);
```
ds-sloth, line 465: I need to update the 25 chars per line constant, based on the number of keyboard columns.

ds-sloth (commented):
I looked at everything.

Obviously I’m not expecting TTFs to work on our 16MB target but I would like to get rasters working at least without a significant RAM impact, and I think we can do that. The main steps are to reduce code size by removing any unused or refactoring rarely used functions, and to reduce dynamic RAM usage by reducing the size of the RasChar struct.

This is very exciting and I’m looking forward to getting it merged soon.

Do you want to do localization in the same PR or in another one? I wouldn’t be opposed to targeting 1.3.7 for full localization support and 1.3.6.1 for font rendering.

I can help refactor your code to support char buffers with length if you’d be willing to make that change.

Is it possible for FreeType to fallback to a system font? That might be nice in certain cases. Similarly, is it possible to require no antialiasing from FreeType?

src/fontman/font_manager.cpp
```diff
@@ -0,0 +1,433 @@
...
+#ifdef THEXTECH_ENABLE_TTF_SUPPORT
+static VPtrList<TtfFont>    g_ttfFonts;
+#endif
+static VPtrList<BaseFontEngine*>    g_anyFonts;
+
+//! Default raster font to render text
+static RasterFont      *g_defaultRasterFont = nullptr;
+
+#ifdef THEXTECH_ENABLE_TTF_SUPPORT
+//! Default raster font to render text
```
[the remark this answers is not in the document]
Wohlstand: Ouch, will fix soon

src/control/keyboard.cpp
```diff
@@ -313,6 +313,7 @@ InputMethodProfile_Keyboard::InputMethodProfile_Keyboard()
 #endif
     this->m_hotkeys[Hotkeys::Buttons::Screenshot] = SDL_SCANCODE_F12;
     this->m_hotkeys2[Hotkeys::Buttons::Screenshot] = SDL_SCANCODE_F2;
+    this->m_hotkeys[Hotkeys::Buttons::ToggleFontRender] = SDL_SCANCODE_F4;
```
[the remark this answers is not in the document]
Wohlstand: F keys are presented at every keyboard on the world (except of several models of MacBook Pro that replaced the row of F keys with a touch bar)

src/fontman/font_manager.cpp
```diff
@@ -0,0 +1,433 @@
...
+            {
+                registerFont(&tf);
+                // Set the default TTF font (as a fallback)
+                if(!g_defaultTtfFont)
+                    g_defaultTtfFont = &tf;
+            }
+        }
+    }
+
+    if(overrider.beginGroup("font-overrides"))
```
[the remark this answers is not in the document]
Wohlstand: My mistake... It doesn't depends on TTF, as it allows to assig any loaded font.

src/fontman/font_manager.cpp
```diff
@@ -0,0 +1,433 @@
...
+    g_ttfFonts.clear();
+#endif
+    g_rasterFonts.clear();
+#ifdef THEXTECH_ENABLE_TTF_SUPPORT
+    closeFreeType();
+#endif
+}
+
+
+PGE_Size FontManager::textSize(const std::string& text, int fontID,
```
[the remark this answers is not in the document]
Wohlstand: Yes, it's a multi-line block size measure call, I use it at Moondust Engine to print mesaage boxes and define the size of the box according to this function.

src/fontman/raster_font.cpp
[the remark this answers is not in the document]
Wohlstand: I should take a look here...

src/fontman/raster_font.cpp
Wohlstand: Good advice, I'll try that.

src/fontman/ttf_font.cpp
Wohlstand: Will add.

src/graphics/gfx_message.cpp
Wohlstand: Ye, I removed that behavior

src/graphics/gfx_print.cpp
Wohlstand: Also, string views are not compatible to C++11 which will make game being not buildable on older compilers (currently the minimum compiler version is GCC 4.9, didn't tested on GCC 4.8, so, don't know). But possible to make our own. I want to update these parts too, currently having this as a temporary solution.

Wohlstand:
About system fonts: they are absolutely inconsistent across platforms, but is possible:
- on Windows, check the registry for one of the default font file, and load some pile of font files to make a "database of glyphs"
- on Linux the fontconfig library can be used to find such system fonts
- on macOS, don't know, I shall check the documentation
- on Android too

src/graphics/gfx_print.cpp
ds-sloth: Okay, then I think let's just use sized buffers. Would you like me to implement those call signatures, or you?

src/control/keyboard.cpp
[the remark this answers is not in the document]
ds-sloth: Sorry, I meant, let's not bind it to ANY key by default. Because this is a very rare case for a user to test.

src/fontman/font_manager.cpp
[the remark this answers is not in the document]
ds-sloth: I'll try this myself now -- I really want to test raster fonts on Wii.

src/control/keyboard.cpp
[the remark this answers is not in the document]
Wohlstand: Oh, I got what you meant... that font engine toggler I made for OUR TEST purpose, just to debug that, compare the difference (alignment, offsets, etc.), and I shall remove this functionality completely once we have sure we don't need an old font engine anymore (keep it as a fallback only).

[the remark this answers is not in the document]
Wohlstand: I added the `#ifdef DEBUG_BUILD` to cover that case.

src/fontman/raster_font.cpp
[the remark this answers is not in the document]
ds-sloth: Good point. I think we could certainly remove these t/b/l/r which seem to be intermediates leftover from Moondust, and if we want to minimize RAM we could also simply do grid-index (int16_t), width (uint8_t), and tex-index (uint8_t) which could also indicate valid/invalid.
I do care about this because if we want to support Latin, Cyrillic, and Kana (not even Hangul or Hanzi/Kanji/Hanja), then the current data structure alone would take up at least 8KB.

src/fontman/font_manager.cpp
```diff
@@ -0,0 +1,433 @@
...
+
+int FontManager::fontIdFromSmbxFont(int font)
+{
+    if(font >= c_smbxFontsMapMax || font < 0)
+        return -1;
+    return s_smbxFontsMap[font];
+}
+
+
+void *FontManager::getDefaultTtfFont()
```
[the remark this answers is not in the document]
Wohlstand: Fixed

src/fontman/font_manager.cpp
```diff
@@ -0,0 +1,433 @@
...
+        {
+            pLogWarning("Failed to load the font %s", fontPath.c_str());
+            g_rasterFonts.pop_back();
+        }
+        else   //Register font name in a table
+            registerFont(&rf);
+    }
+
+    if(!g_rasterFonts.empty())
+        g_defaultRasterFont = &g_rasterFonts.front();
```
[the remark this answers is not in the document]
Wohlstand: Returned from ALPHABETICAL order (you'll find the "std::sort()" call that sorts the file system return alphabetically)

src/fontman/font_manager.cpp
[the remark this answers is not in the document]
Wohlstand: Fixed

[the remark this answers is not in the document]
Wohlstand: For your convenience, I sent that fix just now, I'll fix other notices now.

src/fontman/font_manager.cpp
```diff
@@ -0,0 +1,433 @@
...
+                g_ttfFonts.pop_back();
+            }
+            else   //Register font name in a table
+            {
+                registerFont(&tf);
+                // Set the default TTF font (as a fallback)
+                if(!g_defaultTtfFont)
+                    g_defaultTtfFont = &tf;
+            }
+        }
```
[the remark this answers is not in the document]
Wohlstand: Ouch, yes, thanks for a catch!

src/fontman/raster_font.cpp
```diff
@@ -340,7 +340,7 @@ PGE_Size RasterFont::textSize(const std::string &text, uint32_t max_line_lenght,
             if(need_a_ttf)
             {
 #ifdef THEXTECH_ENABLE_TTF_SUPPORT
-                TtfFont *font = reinterpret_cast<TtfFont*>(FontManager::getDefaultTtfFont());
+                TtfFont *font = dynamic_cast<TtfFont*>(FontManager::getDefaultTtfFont());
```
ds-sloth, line 343: Nice, this is much safer :)

src/fontman/font_manager.cpp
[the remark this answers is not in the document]
Wohlstand: Documented that

src/fontman/raster_font.cpp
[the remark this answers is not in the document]
Wohlstand: I'll just try to forbid copying, I'll see what happen...

src/fontman/raster_font.cpp
```diff
@@ -0,0 +1,490 @@
...
+         * 3 - padding right [for non-mono fonts]
+        */
+        if(charX.empty())
+            continue;
+
+        std::u32string ucharX = std_to_utf32(charX);
+        char32_t ch = ucharX[0];
+        //qDebug()<<"=char=" << ch << "=id="<<charPosX.toInt()<<charPosY.toInt()<<"=";
+        RasChar rch;
+#ifndef DEBUG_BUILD
```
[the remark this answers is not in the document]
Wohlstand: Yes, at debug builds this case is better to crash, basically to let the debugger catch the moment and see the full stack state at the moment of a crash.

src/fontman/raster_font.cpp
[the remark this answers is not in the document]
Wohlstand: Okay, I took a small look on this: I added these left/top/right/bottom to simplify don't re-compute glyph coordinates for the needs of OpenGL. I guess I can easily get rid of these fields with no harm.

src/fontman/raster_font.cpp
Wohlstand: Done

src/fontman/ttf_font.cpp
Wohlstand: Done:
```cpp
    uint8_t *image = nullptr;

    try
    {
        image = new uint8_t[4 * width * height];
    }
    catch(const std::bad_alloc &e)
    {
        pLogCritical("TtfFont::TheGlyph: Out of memory: %s", e.what());
        return dummyGlyph;
    }
```

src/globals.h
Wohlstand: Done

src/graphics/gfx_message.cpp
Wohlstand: I optimised this just now, please review.

src/graphics/gfx_message.cpp
Wohlstand: Done it.

src/fontman/font_manager.cpp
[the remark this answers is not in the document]
ds-sloth: Raster fonts work on Wii now! :star2:

src/graphics/gfx_print.cpp
Wohlstand: Just now I checked this and I reminded: this MESS I made as a lazy workaround because I didn't made the "textSize" that accepts the char* pointer and the size value, I going to definitely update this to avoid having such mess in future.

src/graphics/gfx_print.cpp
Wohlstand: Do you mean to make a kind of:
```cpp
int dFont = NewFontRender ? FontManager::fontIdFromSmbxFont(Font) : -1;
```
?

ds-sloth:
All changes look good. I think the editor versions might recreate the map every frame but performance in the editor is not as high of a priority.

src/graphics/gfx_print.cpp
Wohlstand: Okay, I completely turned the whole into const char* and size_t pair everywhere, should be better now.

src/graphics/gfx_print.cpp
Wohlstand: It's an OLD font engine's size measurer. However, because of a mistake, I hadn't added the forwarding into the new font engine. I did it just now.

src/graphics/gfx_print.cpp
Wohlstand: Done.

src/fontman/font_engine_base.h
```diff
@@ -0,0 +1,46 @@
...
+public:
+    virtual ~BaseFontEngine();
+
+    virtual PGE_Size textSize(const std::string &text,
+                              uint32_t max_line_lenght = 0,
+                              bool cut = false, uint32_t fontSize = 14) = 0;
+
+    virtual void printText(const std::string &text,
+                           int32_t x, int32_t y,
+                           float Red = 1.f, float Green = 1.f, float Blue = 1.f, float Alpha = 1.f,
```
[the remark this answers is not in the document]
Wohlstand: That will need to update a whole renderer. But seems, I could apply that. I.e. replacing the 4xfloat with an uint32_t that uses hexidecial 0xRGBA. That's a separate work.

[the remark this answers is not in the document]
Wohlstand: Done https://github.com/Wohlstand/TheXTech/issues/453

src/globals.h
```diff
@@ -1094,6 +1096,8 @@ extern bool resChanged;
 
 //Public MessageText As String 'when talking to an npc
 extern std::string MessageText;
+typedef std::vector<const char*> UTF8CharMap_t;
```
ds-sloth, line 1099: Please document

ds-sloth (approved):
I'm happy to merge (whenever you fix the build on the most popular desktop platforms :) )

ds-sloth:
Looks like the command-line build *with* SDL is currently broken because it is now looking for a shared SDL2 library that isn't getting built.

CLI build without SDL now working fine.

Wohlstand:
@LoveBodhi, good news for you: Windows builds has been fixed!

Wohlstand:
Anyway, a little note: soon I'll update the format of example fonts, so, I'll send them to you soon to let you test some on your end independently.

src/fontman/ttf_font.cpp
```diff
@@ -79,17 +79,24 @@ bool TtfFont::loadFont(const std::string &path)
 {
     SDL_assert_release(g_ft);
     FT_Error error = FT_New_Face(g_ft, path.c_str(), 0, &m_face);
-    SDL_assert(error == 0);
     if(error)
+    {
+        pLogWarning("Failed to load the font: %s", FT_Error_String(error));
```
ds-sloth, line 84: Nice, I was hoping we could do something like this but didn’t know whether FreeType had error strings.

src/fontman/ttf_font.cpp
```diff
@@ -0,0 +1,594 @@
...
+                    {
+                        if(block & (1 << (7 - x)))
+                            dest[x] = 0xFFFFFFFF;
+                        else
+                            dest[x] = 0;
+                    }
+                }
+            }
+        }
+        else if(bitmap.pitch >= 0)
```
ds-sloth, line 494: Wrong condition

src/fontman/ttf_font.cpp
```diff
@@ -0,0 +1,594 @@
...
+        {
+            for(uint32_t b = 0; b < bsize; ++b)
```
ds-sloth, line 496: I just realized all of these are going against memory order; probably more cache friendly to outer loop over row and inner loop over col/block.

ds-sloth (changes requested):

src/fontman/ttf_font.cpp
Wohlstand: I also found that it's need to enable them at build, otherwise they doesn't work by default.

src/fontman/ttf_font.cpp
Wohlstand: I fixed that

src/globals.h
Wohlstand: Done.

src/core/sdl/render_sdl.cpp
```diff
@@ -400,6 +400,15 @@ void RenderSDL::loadTexture(StdPicture &target, uint32_t width, uint32_t height,
...
     target.inited = true;
+    target.w = width;
+    target.h = height;
+    target.frame_w = width;
+    target.frame_h = height;
+
+    target.l.w_orig = width;
+    target.l.h_orig = height;
+    target.l.w_scale = 1.0f;
+    target.l.h_scale = 1.0f;
```
Wohlstand, line 411: When the original size is matched to the actual, the w_orig and h_orig must be zeroes. This logic is used to use a simplified coordinates formula for such cases in the rendering code. When you don't need to use the scaling feature, just set Orig width and height being zeroes.

ds-sloth: I see, so we should keep `w_orig` 0 unless it's a 1x texture. Then, if it's a 1x texture, we need to separately handle the cases where the original texture loader set `w_orig == 0` and `w_orig != 0`.

ds-sloth: Do you want me to fix this or will you?

ds-sloth: Interesting, well, as long as keeping them as defaults is a valid state, then let's not set them in loadTexture. But I do want the outcome following loadTexture to be a fully valid state (there should be no need to set anything externally if it's a 2x sprite, and if it's a 1x sprite, loadTexture_1x should have the proper outcome). Otherwise, it will be difficult to maintain everything in the future.

Wohlstand: Or, do you think we should move the scaling code into the load texture directly?

ds-sloth: I think it would be reasonable to move the scaling code into loadTexture, since it is an implementation detail the loader shouldn't need to think about. But, the signature of loadTexture shouldn't use FI_Bitmaps, and writing our own scaling function for non-Po2 scaling would be nontrivial.

ds-sloth: Maybe this is a feature we can do later, outside of this PR, whenever we refactor the loading and rendering code.

Wohlstand: I think it's will be enough to keep the have the power-of-two scale function as it should cover our needs. Backgrounds would be fine to be downscaled in a such way. But right, do that at another PR. I'll try to ensure that Wii and 3DS cases aren't broken.

Wohlstand: Okay, at least, removal of these lines doesn't matter for 3DS and Wii.

ds-sloth (approved):
I've tested the raster fonts on the various ports and they work very well. I'm happy for us to merge now or whenever you feel like it's ready.

There are a few minor issues left if we want to polish the engine further:
* Specific TTF fallback per font
* Better TTF/raster hybrid rendering (retain offsets)
* Better mixing of raster fonts of different sizes

LoveBodhi:
Should add "TTF Border Colour" as well!
Scr_2023-01-06_11-39-20

src/fontman/font_manager.cpp
```diff
@@ -92,6 +103,26 @@ BaseFontEngine *FontManager::getDefaultTtfFont()
...
+{
+#ifdef THEXTECH_ENABLE_TTF_SUPPORT
+    if(fontName.empty())
+        return nullptr;
+
+    FontsHash::iterator i = g_fontNameToId.find(fontName);
+    if(i == g_fontNameToId.end())
+        return FontManager::getDefaultTtfFont();
+
+    if(g_anyFonts[i->second]->getFontType() == BaseFontEngine::FONT_TTF) // Ensure requested font IS TTF
```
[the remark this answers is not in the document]
Wohlstand: Then I should change the return type of this call. But seems, reasonable.

src/fontman/raster_font.cpp
```diff
@@ -72,7 +73,10 @@ void RasterFont::loadFont(std::string font_ini)
     font.beginGroup("font");
     font.read("tables", tables, 0);
     font.read("name", m_fontName, m_fontName);
-    font.read("ttf-borders", m_ttfBorders, false);
+    font.read("ttf-outlines", m_ttfOutlines, false);
+    font.read("ttf-borders", m_ttfOutlines, m_ttfOutlines); // Alias, Deprecated
```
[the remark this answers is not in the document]
Wohlstand: Then I should ensure that all font files don't use this deprecated alias.

src/fontman/raster_font.cpp
```diff
@@ -351,21 +356,18 @@ PGE_Size RasterFont::textSize(const char* text, size_t text_size, uint32_t max_l
...
+                TtfFont *font = dynamic_cast<TtfFont*>(FontManager::getTtfFontByName(m_ttfFallback));
                 if(font)
                 {
-                    uint32_t font_size_use = m_letterWidth;
+                    uint32_t font_size_use = m_ttfSize > 0 ? m_ttfSize : m_letterWidth;
 
-                    int y_offset = 0;
+                    // int y_offset = 0; // UNUSED
 
                     bool doublePixel = false;
```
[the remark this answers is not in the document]
ds-sloth: To explain this one further, the issue is that currently, doublePixel is only set if font->bitmapSize() is set. But it should be set based on font->doublePixel() if font->bitmapSize() is unset.

Wohlstand:
Okay, from my side, all current notices have been resolved.

PR #450, opened 2023-01-01
1.3.7 performance updates (plus a few bugfixes to cherry-pick)

ds-sloth:
This PR contains a number of performance updates, I'll list the most important ones here:

* Maintains queues of the important NPC traits (active, killed, has unchecked variables, has one of the onscreen reset flags set) and uses them to avoid iterating over the entire list of NPCs every frame
* Maintains a table of the locations of the NPCs and queries it for NPC-NPC collisions and tempBlock collisions
* Adds NPC tempBlocks to a separate table whenever their location does not match the NPC's location
* Maintains a queue of the onscreen blocks/BGOs that is only updated when the layers move enough, the camera scrolls enough, or a block/BGO is added

I found 1-3 vanilla bugs when making these changes; they are tagged with TODO comments. I also found a bug I introduced when updating the YoshiPound code for the block quadtree; this one is already fixed (1c360a83731968077220fbe42be3ae4007c1aa3e) and should be cherrypicked for 1.3.6.1 also.


ds-sloth:
There may be a minor issue with spin block rendering. I'll check this out tomorrow.

Wohlstand (approved):
After I checked the stuff and tested it on my end at various places. Seems fine. We can continue further testing of the things at the `main` branch normally.

Issue #360, opened 2022-07-26
[Enhancement] Add "holytrinity" and "theessentials" Cheats from SMBX2


0lhi:
*holytrinity* triggers:

* jumpman / ahippinandahoppin
* godmode / donthurtme
* shadowstar

*theessentials* triggers all of the above plus:

* (sonics)tooslow

Will make debugging more convenient.

PR #452, opened 2023-01-03
Add a few new cheats (From X2/Other stuff)

SpencerEverly:
I have decided that I'm gonna separate work between branches for certain features. Here's one that adds additional cheats from SMBX2.

List of cheats added:

- supermario200 (Add the max amount of players into the level)
- getmeouttahere (Exits the level without winning)
- foundmycarkey (Instantly grants an keyhole exit)
- mylifegoals (Instantly grants a SMW goal tape exit)
- mysteryball (Instantly grants a SMB3 goal orb exit)
- itsvegas (Instantly grants a SMB3 roulette exit)
- newleaf (Deactivates all active cheats)
- holytrinity (Activates shadowstar, donthurtme, jumpman)
- passerby (Same as above)
- theessentials (Activates sonicstooslow, shadowstar, donthurtme, jumpman)
- theessenjls (Same as above)

In the future hopefully more cheats will be added. For now, this is what I came up with to add.


0lhi:
> foundmycarkeys (Instantly grants an keyhole exit)

I saw this requires to change "*carkeys*" back to "*wherearemycarkeys*" to avoid a conflict. Maybe it could be named `foundmycarkey` or `foundmycar` instead? That's still pretty close to X2, but wouldn't require to remove an established Code.

src/main/cheat_code.cpp
```diff
@@ -2219,7 +2326,7 @@ static void processCheats()
 
     if(cheated)
     {
-        pLogDebug("Cheating detected!!! [%s]\n", oldString.c_str());
+        pLogDebug("Cheating detected!!!! [%s]\n", oldString.c_str());
```
Wohlstand, line 2329: Why four exclamation signs?

Wohlstand (approved):
For me looks fine.

ds-sloth:
These look useful and should resolve #360 nicely.

Are the implementations of the cheats (especially exit cheats) consistent with the SMBX2 implementations? I'm not familiar with how X2 does them. (Does anyone have a link to those implementations?)

Some minor changes: (1) I don't think you need to include compat.h anymore, and (2) you should rename the foundmycarkeys function to foundmycarkey to match the adjusted cheat name.

ds-sloth:
I'm looking at the SMBX2 implementation of the cheats now and yours are somewhat different, especially for the level exit cheats. Let me try comparing X2 and your build to understand better...

0lhi:
> * th**eess**entials (Activates sonicstooslow, shadowstar, donthurtme, jumpman)
> 
> * th**eess**enjls (Same as above)

The "*eess*" can make this code a bit tricky to read/type, especially for international users. I suggest removing the "*the*".   
Additionally, I'd like to have "*holyfour*" as an alias for this.

> mylifegoals (Instantly grants a SMW goal tape exit)

Can the "*my*" be removed? Or is there a chance that could cause a conflict in the future?

> @Wohlstand  For me looks fine.

I want to manually check every addition before merge. Will likely to so at the end of the week.

> @ds-sloth  I'm looking at the SMBX2 implementation of the cheats now and yours are somewhat different, especially for the level exit cheats. Let me try comparing X2 and your build to understand better...

Eager to see what you'll find! For the combined debug cheats, I found that the X2 implementation was flawed (one thing I remember is that shadowstar would often randomly deactivate), so I'm not sure full accuracy to X2 is desirable here.

0lhi:
Btw @SpencerEverly I haven't tried yet, but looking at your code it seems that holytrinity + theessentials only inverts the current state? If my guess is correct, you should make sure that it always either enables all or disables all.

Wohlstand:
I'll trigger the CI to run the build, so, you could test this out.

Wohlstand:
> Btw @SpencerEverly I haven't tried yet, but looking at your code it seems that holytrinity + theessentials only inverts the current state? If my guess is correct, you should make sure that it always either enables all or disables all.

Right, so, before executing a certain cheat call, ensure that its variable is false.

ds-sloth:
I really don't think we need new globals for these. Instead, you can just intelligently check `(ShadowMode || GodMode || MultiHop)`.

ds-sloth:
I've gotten a current SMBX2 build running on wine (finally) and have some information to report.

As @0lhi said, holytrinity and theessentials are broken there and we should try to make an improved implementation where they are not unpredictably turned off when leaving levels. But I don't think we need to add global variables to accomplish this -- it can be done directly using the existing variables.

The level exit codes are more interesting. The X2 versions actually spawn the level exit NPC in question. The ones implemented here just call the appropriate macros. Differences include: (1) no kill of nearby NPCs, (2) no bonus from roulette, (3) no visible goal tape, and (4) the key code does not search for a keyhole and teleport the player to it with a key. These are very large differences, but maybe not a problem if their main purpose is just to unlock the appropriate exits. I don't know enough about X2 to say, but maybe @0lhi would have some input here.

src/main/cheat_code.cpp
```diff
@@ -1775,7 +1775,7 @@ static void holyTrinity()
         NewEffect(10, tempLocation);
     }
     
-    if HolyTrinity
+    if (HolyTrinity)
```
Wohlstand, line 1778: Little note: in the project's code style we don't have the space between if and paren. So, you should write:
```cpp
if(HolyTrinity)
```
Please follow the code style of projects that you contribute.

0lhi:
> @ds-sloth I really don't think we need new globals for these. Instead, you can just intelligently check (ShadowMode || GodMode || MultiHop). (Edit: probably better to use &&.)

> @ds-sloth But I don't think we need to add global variables to accomplish this -- it can be done directly using the existing variables.

Seconding this. An additional `HolyTrinity` variable will introduce inconsistent behavior.

Example: Typing "godmode", "jumpman", and "shadowstar" enables them. Then typing "holytrinity" plays the enabled sound even though they are already all enabled. Tying "holytrinity" once more will disable them again. If someone wants to enable only "godmode" now it has to be typed twice.

@SpencerEverly "holytrinity" should disable all **if** the `GodMode`, `MultiHop` ***and*** `ShadowMode` bools are ***all true*** - **else** enable all.

> @ds-sloth I don't know enough about X2 to say, but maybe @0lhi would have some input here.

Don't have any experience with the exit codes yet, will share in a day or two after I tested.

src/main/cheat_code.cpp
```diff
@@ -1775,7 +1773,7 @@ static void holyTrinity()
         NewEffect(10, tempLocation);
     }
     
-    if(HolyTrinity)
+    if(!ShadowMode && !GodMode && !MultiHop)
```
ds-sloth, line 1776: We were saying in the discussion that it would be preferable for holytrinity to enable if *any* of the three were disabled, not if all of the three were.

You may also want to play the grow/shrink sound inside of the subclauses.

Wohlstand: So, the logic must be:
- When ANY is enabled, enable every other disabled.
- When ALL are enabled, disable them all.
```cpp
if(ShadowMode && GodMode && MultiHop)
{
   // Disable everything
}
else
{
    if(!ShadowMode)
    {
        // enable it
    }

    if(!GodMode)
    {
        // enable it
    }

    if(!MultiHop)
    {
        // enable it
    }
}
```

0lhi: Wouldn't it suffice to do

```c++
if(ShadowMode && GodMode && MultiHop)
{
   // Disable everything
}
else
{
   // Enable everything
}
```

?

Or is there an advantage for the additional check that I'm not seeing?

Wohlstand: I guess, for such an effect, we better to have one single sound, no need to play the sound of each effect trice :eyes:

P-Star7:
"theessenjls" might be a bad idea as it might imply that TheXTech has involvement with or an endorsement from Enjl, who has nothing to do with this project.

SpencerEverly:
> "theessenjls" might be a bad idea as it might imply that TheXTech has involvement with or an endorsement from Enjl, who has nothing to do with this project.

Oh yeaaaah, that's because I got the cheats list from https://docs.codehaus.moe/#/features/cheats , that may be why I preserved that as well

0lhi:
Some comments about each Exit Cheat:

`foundmycarkey/s`: In X2 it only works when there actually is a secret exit (player gets teleported there), while in TheXTech-SEE it works always. For debug purposes the latter approach is superior.

SMBX2: 


TheXTech-SEE:
Scr_2023-01-07_17-01-18


We should make sure that the new keyhole effect #458 will be visible when the Cheat is toggled.



`mysteryball`: In practise works the same, except for the minor detail that enemies aren't killed off (also true for the next ones). Since the character is invincible during the animation, this is fine. In some cases it is superior.

SMBX2:



TheXTech-SEE:
Scr_2023-01-07_16-21-39


`mylifegoals`: Again, in practice mostly the same. The Flagpole Sprite in X2 is rather ugly and immersion breaking. I consider it's absence in TheXTech-SEE to be an advantage...



Player despawns after a few seconds no matter what, which is good. However, the same doesn't apply to~

`itsvegas`: On both, X2 and TheXTech-SEE it is super easy to get softlocked.



We need to do better. I see two possible approaches:

1. Add the fadeout from the flagpole exit on the condition that the cheat has been triggered.

2. Kill off the player a few seconds after the cheat has been triggered if the level wasn't left yet. Maybe add a visual or audible indicator for the coming death. This idea may be more fitting since it adds a gambling component.

Overall I think Spencer's approach is safer and more reliable compared to SMBX2, so we should go with it once the remaining issues are fixed.

0lhi:
I addressed some of the Issues I mentioned in the recent commit. @ds-sloth @Wohlstand Can you take a look at my changes?

If it's okay the only thing missing for Merge is the Keyhole effect.

ds-sloth:
The base code looks promising and I appreciate 0lhi's logic fixes.

I made a few additions:
(1) merged the main branch in for testing
(2) made some quick style fixes
(3) modified the end condition of the softlock case to allow the full fade animation to take place
(4) partially spawned a BGO for the keyhole animation feature. It's a bit tricky to properly spawn BGOs because they are normally sorted in a particular order
(5) updated the changelog

I'm fine to merge as-is, but I'd prefer to avoid adding the global variable `ExitCheat` for this relatively small feature. I'd suggest we could re-use the int `LevelMacroWhich` to track the case where the roulette cheat is triggered, and replace the guard `if(ExitCheat) return;` with `if(LevelMacro != LEVEL_MACRO_NONE) return;`. What do the rest of you think?

0lhi:
> and replace the guard `if(ExitCheat) return;` with `if(LevelMacro != LEVEL_MACRO_NONE) return;`

Can you make it so that these two conditions are met?

1. Exit Cheats cannot be activated while another Exit Cheat is being executed.
2. Exit Cheats can be used to escape Softlocks caused my misplaced Roulette NPC.

ds-sloth:
> Can you make it so that these two conditions are met?

Sure, that's easy enough.

(1) This would be safe to do by checking LevelMacro for everything but itsvegas.

(2) We'd add a special condition for itsvegas where it could trigger during a level macro so long as it wasn't originally triggered by a cheat (could still use LevelMacroWhich to track), and then check for that value of LevelMacroWhich in the level macro code where you currently check for ExitCheat.

Does that sound reasonable to you?

0lhi:
> Does that sound reasonable to you?

Yes! Gonna test once you commit =)

0lhi (approved):
Good for Merge =)

ds-sloth:
Alright, I'm happy to merge also whenever @Wohlstand is.

Wohlstand:
I take a quick look, and ye, I don't see anything suspicious right now, so, it's fine to merge.

PR #458, opened 2023-01-05
Keyhole animation ready for testing

ds-sloth:
Should be a quick test but wanted to do it via a PR so the implementation could be reviewed. How does it look?


Wohlstand (commented):
By a quick look, seems fine, I should check this more carefully sooner.

ds-sloth:
Before we merge, I just realized: (1) there isn't a compat guard or changelog entry for the animation, and (2) the animation doesn't work perfectly when the keyhole framerate fix is disabled.

0lhi:
I think for now you can just disable the animation when `fix-keyhole-framerate = false`, no reason to introduce another flag for this.

Wohlstand:
> the animation doesn't work perfectly when the keyhole framerate fix is disabled.

You should use the floating iterator for that. That's a way to perform a variadic-framerate update.

ds-sloth:
I thought I was already properly adjusting the framerate. It's probably a silly fix -- I'll take a look now.

Anyway, I wonder if we should just re-use the keyhole framerate flag and decide which flag to use permanently when merging the config branch in the future.

ds-sloth:
Fixed that the !fix_keyhole_framerate issue -- turns out that the entire UpdateGraphics call was being cancelled by a frameskip. The fix_keyhole_framerate workaround didn't update the fields used by the new frameskip code. So I disabled frameskip during the keyhole exit macro if the fix_keyhole_framerate was off.

I also added documentation, and a compat guard for the feature. So I think we are now ready to go if the code looks good to you.

Wohlstand (approved):
I checked the thing, and I made some small tweaks at the gfx_update to simplify the condition. So, I going to merge this.

PR #466, opened 2023-01-10
Keep old gamesave file at episode directory

### LINKED ISSUE #436: Feature Request: Keep old Save File upon deletion; make empty savx instead
The current behavior is that the old Vanilla *save1.sav* gets removed from the Episode Folder. This is undesirable behavior, as the content inside there shouldn't under any circumstances be changed outside of the Editor.

Instead, an empty file should be made in the gamesave directory that would tell TheXTech to ignore the Vanilla .sav file.

Wohlstand:
Make a full sure that the episode directory will be absolutely read-only-friendly even if an old gamesave file was left in it. Instead of actual .sav file deletion, create a locker file that makes gamesave to be always empty. Once gamesave gets created, then remove that locker file.

Also, remove the support for SAVX files at the episode directory: this is fully deprecated behaviour that existed in early TheXTech versions only and is no longer used for a long time.

Fixes #436

### DIFF
--- a/src/game_main.cpp
+++ b/src/game_main.cpp
@@ -1933,17 +1933,12 @@ void DeleteSave(int world, int save)
     deleteList.push_back(makeGameSavePath(w.WorldPath, \
                                           w.WorldFile,\
                                           fmt::format_ne(f, save)));
-#define AddFileW(f) \
-    deleteList.push_back(w.WorldPath + fmt::format_ne(f, save));
 
     AddFile("save{0}.savx");
     AddFile("timers{0}.ini");
     AddFile("deaths-{0}.rip");
     AddFile("fails-{0}.rip");
     AddFile("demos-{0}.dmo");
-    // Old gamesaves
-    AddFileW("save{0}.savx");
-    AddFileW("save{0}.sav");
 
     // Clear all files in list
     for(auto &s : deleteList)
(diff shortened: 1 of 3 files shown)


ds-sloth:
I read the code and it looks clean to me. I haven't tested it yet.

One small note: it looks like existence of the nosave will prevent a **Modern** save file from being loaded too. There is a danger that a save file could be overwritten. I could do a quick test to confirm this. I think we'd just add the locker as an intermediate condition between the two current conditions.

ds-sloth:
It's not hard to get a situation where both a `nosave` and a modern `savx` exist for the same slot, even with operating system corruption: (1) load episode with legacy save in slot 1, (2) delete legacy save (slot 1), (3) copy modern save from slot 2 to slot 1.

But, everything works well in my testing after your most recent commit, so I am happy to approve for merge. It would be nice to see @0lhi test with some archival episodes also.

ds-sloth (approved):

0lhi:
Looks good to me. Couldn't make a savx and nosave exist at the same time through the game, and when I made them both exist manually, it will still load the new safe. Only suggestion I have is to make the nosave text a bit more elaborate:

```
If this file exists, the Vanilla "save3.sav" inside the episode folder is ignored.
```

0lhi (approved):

Wohlstand:
I'll apply the string update later as I'm out of my home. @ds-sloth please help me to apply the string update if you can, I'm out of home now, can't my myself now.

PR #469, opened 2023-01-15
Misc source code improvements + fixes for constants

ds-sloth:
I noticed that the multires PR has too many changes to be easy to review, and wanted to move some of the "easy" changes out of it so that we can discuss it in stages.

This PR only includes the safest changes (general code cleanup, replacing screen-size menu and HUD with variables based on ScreenW/ScreenH).

There are also a few bugfixes (mostly input-related) that I included only at the multires branch; that's simply a mistake. I included them here too.

And one feature (MagicHand snaps to layer grid if the player is level-testing with a moving layer).

I hope it should be easy to review, and simplify the later review of the other multires changes.

Note: this should be safe for merge before 1.3.6.1. The only user-facing changes are the minor bugfixes discussed above.

### DIFF
--- a/src/graphics.cpp
+++ b/src/graphics.cpp
@@ -74,13 +74,13 @@ void GetvScreen(const int A)
 //    }
 //    else
     {
-        double pHeight = (p.Mount != 2) ? pLoc.Height : 0;
-
+        if(p.Mount == 2)
+            pLoc.Height = 0;
         // this guard is new because players can now respawn in 1P mode through DropAdd
         if(Player[A].Effect != 6)
         {
             vScreenX[A] = -pLoc.X + (vScreen[A].Width * 0.5) - pLoc.Width / 2.0;
-            vScreenY[A] = -pLoc.Y + (vScreen[A].Height * 0.5) - vScreenYOffset - pHeight;
+            vScreenY[A] = -pLoc.Y + (vScreen[A].Height * 0.5) - vScreenYOffset - pLoc.Height;
             vScreenX[A] += -vScreen[A].tempX;
             vScreenY[A] += -vScreen[A].TempY;
             if(-vScreenX[A] < level[p.Section].X)
@@ -105,6 +105,8 @@ void GetvScreen(const int A)
             if(vScreen[A].TempY < 0)
                 vScreen[A].TempY += 1;
         }
+        if(p.Mount == 2)
+            pLoc.Height = 128;
     }
 }
 
--- a/src/graphics.cpp
+++ b/src/graphics.cpp
@@ -74,13 +74,13 @@ void GetvScreen(const int A)
 //    }
 //    else
     {
-        if(p.Mount == 2)
-            pLoc.Height = 0;
+        double pHeight = (p.Mount != 2) ? pLoc.Height : 0;
+
         // this guard is new because players can now respawn in 1P mode through DropAdd
         if(Player[A].Effect != 6)
         {
             vScreenX[A] = -pLoc.X + (vScreen[A].Width * 0.5) - pLoc.Width / 2.0;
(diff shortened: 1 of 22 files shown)


src/graphics.cpp
```diff
@@ -105,8 +105,6 @@ void GetvScreen(const int A)
             if(vScreen[A].TempY < 0)
                 vScreen[A].TempY += 1;
         }
-        if(p.Mount == 2)
-            pLoc.Height = 128;
```
Wohlstand, line 109: Why you removed this assignment?

ds-sloth: The purpose of that commit was to avoid using the player location as a temporary variable, but now I see that it's possible that the player's previous location was not 128. So, I'll revert this and leave a note for it.

ds-sloth: I decided to simply revert the commit. There isn't enough benefit to justify the complication of changing and explaining the case in the code.

src/graphics/gfx_special_frames.cpp
```diff
@@ -116,3 +116,131 @@ void SpecialFrames()
...
+{
+    BackgroundFrameCount[26]++;
+    if(BackgroundFrameCount[26] >= 8)
+    {
+        BackgroundFrame[26]++;
+        if(BackgroundFrame[26] >= 8)
+            BackgroundFrame[26] = 0;
+        BackgroundFrameCount[26] = 0;
+    }
+    BackgroundFrameCount[18]++;
```
Wohlstand, line 129: Please put the space after closing the bracket to don't mess up the overlook. At here, and at other places.

src/main/game_loop.cpp
```diff
@@ -509,6 +509,7 @@ int PauseGame(PauseCode code, int plr)
         Player[i].UnStart = false;
         Player[i].CanJump = false;
     }
+    MenuCursorCanMove = false;
```
Wohlstand, line 513: Empty line after closing bracket wants to be here too :eyes:

src/graphics/gfx_special_frames.cpp
ds-sloth: Good idea, I standardized the whole file.

src/main/menu_main.cpp
```diff
@@ -1661,16 +1691,16 @@ void mainMenuDraw()
...
+            SuperPrint(g_mainMenu.mainBattleGame, 3, MenuX, MenuY+30*(i++));
         if(g_config.enable_editor)
-            SuperPrint(g_mainMenu.mainEditor, 3, 300, 350+30*(i++));
-        SuperPrint(g_mainMenu.mainOptions, 3, 300, 350+30*(i++));
-        SuperPrint(g_mainMenu.mainExit, 3, 300, 350+30*(i++));
-        XRender::renderTexture(300 - 20, 350 + (MenuCursor * 30), 16, 16, GFX.MCursor[0], 0, 0);
+            SuperPrint(g_mainMenu.mainEditor, 3, MenuX, MenuY+30*(i++));
+        SuperPrint(g_mainMenu.mainOptions, 3, MenuX, MenuY+30*(i++));
+        SuperPrint(g_mainMenu.mainExit, 3, MenuX, MenuY+30*(i++));
+        XRender::renderTexture(MenuX - 20, MenuY + (MenuCursor * 30), 16, 16, GFX.MCursor[0], 0, 0);
```
Wohlstand, line 1703: At any arithmetical expression, it's better to always add space between operands and operators. Also, when it's a multiplication and adding `x + y * z`, better to add parens just in case `x + (y * z)`. I do know some faulty compilers that result in the wrong result (some very old GCC, but I don't remind the exact place where I caught this mess...).

Wohlstand (approved):
If not count the minor style notice, the stuff seems fine. There is testing needed to ensure is anything wrong or not.

PR #473, opened 2023-01-20
Show gamesave info on title screen

ds-sloth:
Small PR with just the feature described and supporting infrastructure. I believe this feature has been requested a few times and it seemed simple enough to implement.

 

Major changes:
* SaveSlot and SaveStars are replaced with a new SaveSlotInfo array which contains much more information about the gamesave.
* This information is displayed at the title screen.

Minor changes:
* SavedChar now uses a unique `SavedChar_t` instead of a `Player_t`. This is harmless because the other variables in SavedChar were only written, never read. (Note: this saves more memory than SaveSlotInfo costs.)
* Fixed some inconsistencies in the settings for the Fails counter (it was previously off by default if `gameinfo.ini` existed and on by default otherwise).
* The Fails counter now reads "Fails" unless `gameinfo.ini` requests "Demos".

### DIFF
--- a/src/globals.cpp
+++ b/src/globals.cpp
@@ -305,8 +305,12 @@ int selSave = 0;
 int PSwitchTime = 0;
 int PSwitchStop = 0;
 int PSwitchPlayer = 0;
-RangeArrI<int, 1, 3, 0> SaveSlot;
-RangeArrI<int, 1, 3, 0> SaveStars;
+
+// RangeArrI<int, 1, maxSaveSlots, 0> SaveSlot;
+// RangeArrI<int, 1, maxSaveSlots, 0> SaveStars;
+
+RangeArr<SaveSlotInfo_t, 1, maxSaveSlots> SaveSlotInfo;
+
 int BeltDirection = 0;
 bool BeatTheGame = false;
 //int cycleCount = 0;
@@ -361,7 +365,7 @@ int Score = 0;
 RangeArrI<int, 1, 13, 0> Points;
 int MaxWorldStars = 0;
 bool Debugger = false;
-RangeArr<Player_t, 0, 10> SavedChar;
+RangeArr<SavedChar_t, 0, 10> SavedChar;
 
 bool LoadingInProcess = false;
 int LoadCoins = 0;
--- a/src/globals.cpp
+++ b/src/globals.cpp
@@ -306,9 +306,6 @@ int PSwitchTime = 0;
 int PSwitchStop = 0;
 int PSwitchPlayer = 0;
 
-// RangeArrI<int, 1, maxSaveSlots, 0> SaveSlot;
-// RangeArrI<int, 1, maxSaveSlots, 0> SaveStars;
-
 RangeArr<SaveSlotInfo_t, 1, maxSaveSlots> SaveSlotInfo;
 
 int BeltDirection = 0;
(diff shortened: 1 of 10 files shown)


src/globals.cpp
```diff
@@ -305,8 +305,12 @@ int selSave = 0;
 int PSwitchTime = 0;
 int PSwitchStop = 0;
 int PSwitchPlayer = 0;
-RangeArrI<int, 1, 3, 0> SaveSlot;
-RangeArrI<int, 1, 3, 0> SaveStars;
+
+// RangeArrI<int, 1, maxSaveSlots, 0> SaveSlot;
+// RangeArrI<int, 1, maxSaveSlots, 0> SaveStars;
```
Wohlstand, line 310: Maybe remove these lines? :thinking:

src/main/game_save.cpp
```diff
@@ -51,18 +53,60 @@ std::string makeGameSavePath(std::string episode, std::string world, std::string
     return ret;
 }
 
+static void s_LoadCharacter(SavedChar_t& dest, const saveCharState& s)
+{
+    dest = SavedChar_t();
+
+    if((s.state < 1) || (s.state > 10))
```
Wohlstand, line 60: Maybe put the "max state" constant rather than using the 10 magic number? I mean, for the future case when we add more state slots (for example, SMBX-38A's).

ds-sloth: Very good idea, I think this was used elsewhere in our code but I'll update them both.

Wohlstand (approved):
Seems good in general, but I hadn't tested it in action.

ds-sloth:
@0lhi can you please test the functionality of this at your convenience?

0lhi:
Speedrun-Mode is still flexible, so you should only display the time when the Game has been started in a Speedrun-Mode.

ds-sloth:
Good catch. Currently the time only shows for runs where the speedrun mode was previously active.

Should we warn the user if there's a time saved, but the game isn't launched in speedrun mode? I don't want to make it annoying, but it might be a useful feature. (I remember that in 1.3.7 we're planning to make speedrun a savefile-specific setting.)

LoveBodhi:
Scr_2023-01-23_08-21-05

"Deaths" still works fine.

0lhi:
> Should we warn the user if there's a time saved, but the game isn't launched in speedrun mode?

Nah. Currently time doesn't get tracked outside of Speedrun, so that's expected behavior.

0lhi (approved):

ds-sloth:
Wohlstand, we can merge if you like the functionality when you test it. Otherwise, I'm happy to discuss/make changes.

src/main/menu_main.cpp
```diff
@@ -1560,6 +1562,62 @@ static void s_drawGameSaves()
...
+
+    // Print coins on the screen (from gfx_update2.cpp)
+    int coins_x = infobox_x + 480 - 10 - 36 - 62;
+    XRender::renderTexture(coins_x + 16, row_lc, GFX.Interface[2]);
+    XRender::renderTexture(coins_x + 40, row_lc + 2, GFX.Interface[1]);
+    SuperPrintRightAlign(t = std::to_string(info.Coins), 1, coins_x + 62 + 36, row_lc + 2);
+
+    // Fails Counter
+    if(gEnableDemoCounter)
+        SuperPrintRightAlign(fmt::format_ne("{0}: {1}", gDemoCounterTitle, info.Fails), 3, infobox_x + 480 - 10, row_2);
```
Wohlstand, line 1620: Little note that the demos counter can be enabled at the episode side too via `compat.ini` :eyes:

PR #482, opened 2023-02-02
Fix two vanilla bugs that kill player via the pinched code

ds-sloth:
Developer-facing changes:
- Multiple Pinched values (5 integers) replaced with a single bitfield (`PinchedInfo_t`), with full mememu support. Saves ~100kb of RAM and makes NPCs more cache-friendly. Small CPU expense, but the pinched info isn't accessed in hot loops.

User-facing changes:
- Fixed vanilla bug where player could be prematurely crushed by a sloped ceiling
 

- Fixed vanilla bug where player could be crushed by a horizontally (not vertically) moving ceiling (these examples are from the same gameplay recording script, in which the Mode 3 player dies and the Mode 1 player survives)
 

I'd like to include this in 1.3.6.1, but, we should probably make a gameplay feature freeze soon because the main branch now includes the 1.3.7 performance updates.

[pinch-tests.zip](https://github.com/Wohlstand/TheXTech/files/10574388/pinch-tests.zip)

### DIFF
--- a/src/script/luna/mememu.cpp
+++ b/src/script/luna/mememu.cpp
@@ -1329,11 +1329,14 @@ public:
         insert(0x00000140, &Player_t::Immune);
         insert(0x00000142, &Player_t::Immune2);
         insert(0x00000144, &Player_t::ForceHitSpot3);
-        insert(0x00000146, &Player_t::Pinched1);
-        insert(0x00000148, &Player_t::Pinched2);
-        insert(0x0000014a, &Player_t::Pinched3);
-        insert(0x0000014c, &Player_t::Pinched4);
-        insert(0x0000014e, &Player_t::NPCPinched);
+
+        // handlers below
+        // insert(0x00000146, &Player_t::Pinched1);
+        // insert(0x00000148, &Player_t::Pinched2);
+        // insert(0x0000014a, &Player_t::Pinched3);
+        // insert(0x0000014c, &Player_t::Pinched4);
+        // insert(0x0000014e, &Player_t::NPCPinched);
+
         insert(0x00000150, &Player_t::m2Speed);
         insert(0x00000154, &Player_t::HoldingNPC);
         insert(0x00000156, &Player_t::CanGrabNPCs);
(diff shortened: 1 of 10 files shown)


src/script/luna/mememu.cpp
```diff
@@ -1499,7 +1562,25 @@ class NPCMemory final : public NpcParent
...
+            case 0x0C: // Pinched2
+                return valueToMem((int)obj->Pinched.Left2, ftype);
+            case 0x0E: // Pinched3
+                return valueToMem((int)obj->Pinched.Top3, ftype);
+            case 0x10: // Pinched4
+                return valueToMem((int)obj->Pinched.Right4, ftype);
+            case 0x12: // MovingPinched
+                return valueToMem((int)obj->Pinched.Moving, ftype);
+            default:
+                // use default handler
```
Wohlstand, line 1580: I guess, the warning message should be added here in any cases of attempts to hack bytes in this area.

src/script/luna/mememu.cpp
```diff
@@ -1524,6 +1605,40 @@ class NPCMemory final : public NpcParent
...
+                obj->Pinched.Top3 = in;
+                return;
+            case 0x10: // Pinched4
+                obj->Pinched.Right4 = in;
+                return;
+            case 0x12: // MovingPinched
+                obj->Pinched.Moving = in;
+                return;
+            default:
+                // use default handler
```
Wohlstand, line 1638: Same as above:
I guess, the warning message should be added here in any cases of attempts to hack bytes in this area.

src/script/luna/mememu.cpp
ds-sloth: Good idea, I'll fix that now in both cases

0lhi:
This seems wrong:

Scr_2023-02-03_01-14-37

ds-sloth:
Strangely enough, I think that's actually the Mode 3 behavior too. I'd rather just fix the unexpected deaths, not every bug related to the slope code (which is certainly not ideal).

0lhi:
> Strangely enough, I think that's actually the Mode 3 behavior too. I'd rather just fix the unexpected deaths, not every bug related to the slope code (which is certainly not ideal).

True, I could reproduce in Mode 3 with God Mode :o 

Scr_2023-02-10_04-10-25

I'm not sure how I feel about this. By fixing one issue, another gets more prominence. It might be better to leave it unchanged.

ds-sloth:
It seems like the undesirable behavior is exclusive to the normal blocks moving downwards while the player is on a sloped floor.

When the sloped floor moves upwards, both vanilla and TheXTech allow the player to clip through the ceiling.



I would be happy to revert the changes for the sloped floor case. I imagine it's a lot less common than the sloped ceiling case, and it's also glitchier in the vanilla game already.

ds-sloth:
I reverted the sloped floor case, and also retargeted this change to 1.3.7.

We might want to revisit the behavior of this compat flag before releasing that version if somebody notices another unusual case during testing.

ds-sloth:
@Wohlstand @0lhi does this seem good to merge? We can revise the behavior more later if anything unexpected emerges before the 1.3.7 release, but I think the essential issues have been discussed now.

0lhi:
Will this be excluded from the 1.3.6.1 Release?

Merge should be okay if you make an issue to do more testing before 1.3.7.

ds-sloth:
I would like to exclude this from 1.3.6.1, yes.

Wohlstand:
Seems to be fine to merge, but still need a further testing to ensure nothing significant got broken.

PR #486, opened 2023-02-08
Add new speedless location type

ds-sloth:
I nearly committed straight to main, but it does impact a few objects so I thought it would be nice to get some review.

This adds a new type `SpeedlessLocation_t` without the `SpeedX` or `SpeedY` attributes, and uses it for the `Location` attribute of objects that do not ever use these attributes. Those objects are Warp, Water, the World items, Credit Line, and Event Section.

In order to avoid any mistakes or inefficiencies, I made its conversion functions to/from `Location_t` explicit.

This change saves of 1.4 MB of RAM on the PC build, and 400 KB of RAM on the low-RAM builds. It has no gameplay effects, and none of the affected items had Autocode bindings.

### DIFF
--- a/src/collision.cpp
+++ b/src/collision.cpp
@@ -37,6 +37,51 @@ bool CheckCollision(const Location_t &Loc1, const Location_t &Loc2)
     return false;
 }
 
+bool CheckCollision(const SpeedlessLocation_t &Loc1, const Location_t &Loc2)
+{
+    if(
+        (Loc1.Y + Loc1.Height >= Loc2.Y) &&
+        (Loc1.Y <= Loc2.Y + Loc2.Height) &&
+        (Loc1.X <= Loc2.X + Loc2.Width) &&
+        (Loc1.X + Loc1.Width >= Loc2.X)
+    )
+    {
+        return true;
+    }
+
+    return false;
+}
+
+bool CheckCollision(const Location_t &Loc1, const SpeedlessLocation_t &Loc2)
+{
+    if(
+        (Loc1.Y + Loc1.Height >= Loc2.Y) &&
+        (Loc1.Y <= Loc2.Y + Loc2.Height) &&
+        (Loc1.X <= Loc2.X + Loc2.Width) &&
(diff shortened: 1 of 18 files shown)


src/main/block_table.hpp
```diff
@@ -293,7 +293,7 @@ struct screen_t
 template<class MyRef_t>
 inline Location_t extract_loc(MyRef_t obj)
 {
-    return obj->Location;
+    return (Location_t)obj->Location;
```
Wohlstand, line 296: Maybe use a kind of `static_cast<type>()`?

ds-sloth: There are many such cases, but I can switch them all to static cast if you prefer.

src/collision.cpp
```diff
@@ -37,6 +37,51 @@ bool CheckCollision(const Location_t &Loc1, const Location_t &Loc2)
...
+        (Loc1.Y <= Loc2.Y + Loc2.Height) &&
+        (Loc1.X <= Loc2.X + Loc2.Width) &&
+        (Loc1.X + Loc1.Width >= Loc2.X)
+    )
+    {
+        return true;
+    }
+
+    return false;
+}
```
Wohlstand, line 53: You could simplify this by just returning the result of the condition:
```
return (Loc1.Y + Loc1.Height >= Loc2.Y) &&
       (Loc1.Y <= Loc2.Y + Loc2.Height) &&
       (Loc1.X <= Loc2.X + Loc2.Width) &&
       (Loc1.X + Loc1.Width >= Loc2.X);
```

ds-sloth: Yes, that is superior. I was just copy-pasting each of the collision calls. This could be a good chance to refactor them.

src/main/block_table.hpp
ds-sloth: Done!

src/collision.cpp
ds-sloth: Also done.

Wohlstand (approved):
Looks like it's fine now. I going to merge this.

PR #492, opened 2023-02-18
Fill ZH-CN Translation #4

LoveBodhi:


ds-sloth:
Does 颗 work for Leeks, and other possible Star object names? If not, is 个 acceptable for both Stars and Leeks?

If neither works for both cases, I can try to make the counter word customizable, or you could include the counter word in the localized name for stars / leeks.

LoveBodhi:
Thery're different classifiers in CN, therefore must use both (Stars for 颗, Leeks for 根)

ds-sloth:
Thanks for the clarification!
Could you please include the classifier in the noun's translation, instead of the phrase? I know that isn't strictly accurate, but I will make sure that this translation isn't used in any cases that would require the noun without its counter word.
If many more cases of this occur, we can discuss a general strategy with @Wohlstand at that time.

LoveBodhi:
Customizable CN exclusive counter word should be added as well.

ds-sloth:
To minimize the number of fields, I'll use the same translation field for the dual as well as the counter word, unless we find a language that has both a dual noun form and counter words.

I'll rename that field "DualOrCounter" and move your counter words into it. It will be passed to the phrase in slot 2, as you expected.

ds-sloth:
I'll do some slight rearrangement after merging. Thanks for the translations!

ds-sloth:
I updated the assets and game logic after merging -- please verify that the behavior for locked warps looks correct to you.

PR #499, opened 2023-03-16
Fill ZH-CN Strings

LoveBodhi:


ds-sloth:
Looks great, thanks very much for the hard work!

PR #506, opened 2023-03-28
Compat autoconversion for 1.3.6.1

ds-sloth:
Relatively small PR that:

* Adds a flexible user Prompt screen we can use for multiple purposes
* Adds renameKey and deleteKey methods to IniPorcessing
* Tracks in SelectWorld and SelectBattle whether the path is writable
* ~~Adds config preferences for compat autoconversion and for warning the user if autoconversion fails~~
* Uses all of these to ask the user and automatically convert `compat.ini` files whenever possible. Asks the user about an individual file only once per run.

All of the strings are stored in the main menu class and should be easy to integrate with localization when ready.

Review requested for:
* Logical functionality (do we like only prompting the user once per run? are the options I added good?)
* Asthetics of the Prompt screen
* Technical implementation (from @Wohlstand)
* Any other notes

Example for a file I marked as unwritable (via `chmod -w`):

**Updated** screenshots of the two screens:

### DIFF
--- a/lib/IniProcessor/ini_processing.cpp
+++ b/lib/IniProcessor/ini_processing.cpp
@@ -653,6 +653,43 @@ bool IniProcessing::hasKey(const std::string &keyName) const
     return (e != m_params.currentGroup->end());
 }
 
+bool IniProcessing::renameKey(const std::string &oldName, const std::string &newName)
+{
+    if(!m_params.opened)
+        return false;
+
+    if(!m_params.currentGroup)
+        return false;
+
+    auto it = m_params.currentGroup->find(oldName);
+    if(it != m_params.currentGroup->end())
+    {
+        std::swap((*m_params.currentGroup)[newName], it->second);
+        m_params.currentGroup->erase(it);
+        return true;
+    }
+
+    return false;
+}
+
+bool IniProcessing::deleteKey(const std::string &keyName)
+{
+    if(!m_params.opened)
+        return false;
+
+    if(!m_params.currentGroup)
+        return false;
+
+    auto it = m_params.currentGroup->find(keyName);
+    if(it != m_params.currentGroup->end())
+    {
+        m_params.currentGroup->erase(it);
+        return true;
+    }
+
+    return false;
+}
+
 std::vector<std::string> IniProcessing::allKeys() const
 {
     std::vector<std::string> keys;
--- a/lib/IniProcessor/ini_processing.cpp
+++ b/lib/IniProcessor/ini_processing.cpp
@@ -342,10 +342,10 @@ bool IniProcessing::parseHelper(char *data, size_t size)
                 value = lskip(end + 1);
                 end = find_char_or_comment(value, '\0');
 
-                #ifndef CASE_SENSITIVE_KEYS
+#ifndef CASE_SENSITIVE_KEYS
                 for(char *iter = name; *iter != '\0'; ++iter)
                     *iter = (char)tolower(*iter);
-                #endif
+#endif
 
                 if(*end == ';')
                     *end = '\0';
@@ -359,9 +359,9 @@ bool IniProcessing::parseHelper(char *data, size_t size)
                     if(!recentKeys)
                         recentKeys = &m_params.iniData["General"];
 
-                    #ifdef INIDEBUG
+#ifdef INIDEBUG
                     printf("-> [%s]; %s = %s\n", section, name, v);
-                    #endif
+#endif
                     (*recentKeys)[name] = unescapeString( removeQuotes(v, v + strlen(v)) );
                 }
             }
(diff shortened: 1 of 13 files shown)


lib/IniProcessor/ini_processing.cpp
```diff
@@ -653,6 +653,43 @@ bool IniProcessing::hasKey(const std::string &keyName) const
...
+    auto it = m_params.currentGroup->find(keyName);
+    if(it != m_params.currentGroup->end())
+    {
+        m_params.currentGroup->erase(it);
+        return true;
+    }
+
+    return false;
+}
+
```
Wohlstand, line 692: These things would go into the main repository of IniProcessing too. https://github.com/WohlSoft/IniProcessing

LoveBodhi:
Should we use a customizable background and music for the warning screen?

ds-sloth:
Probably not for this compat question, this screen will only exist for version 1.3.6.1.

However, it might make the general Prompt screen more useful in the long term if we make background and music internal options when calling it. I'd be happy to add that.

lib/IniProcessor/ini_processing.cpp
ds-sloth: Should I commit these changes directly to that repo or submit a PR?

Wohlstand: I'll take care by myself :)
Just a note that this change will be also copied into main repo of module.

lib/IniProcessor/ini_processing.cpp
```diff
@@ -653,6 +653,43 @@ bool IniProcessing::hasKey(const std::string &keyName) const
...
+    if(!m_params.currentGroup)
+        return false;
+
+    auto it = m_params.currentGroup->find(oldName);
+    if(it != m_params.currentGroup->end())
+    {
+        std::swap((*m_params.currentGroup)[newName], it->second);
+        m_params.currentGroup->erase(it);
+        return true;
+    }
```
Wohlstand, line 670: What happen if old name and new name will be equal? :)
Looks suspicious...

Wohlstand: P.S. About case sensitivity, pay attention on this code: https://github.com/Wohlstand/TheXTech/blob/bd7de09274a49d221055a2edc3877934a5310712/lib/IniProcessor/ini_processing.h#L95

lib/IniProcessor/ini_processing.cpp
```diff
@@ -653,6 +653,43 @@ bool IniProcessing::hasKey(const std::string &keyName) const
...
+
+bool IniProcessing::deleteKey(const std::string &keyName)
+{
+    if(!m_params.opened)
+        return false;
+
+    if(!m_params.currentGroup)
+        return false;
+
+    auto it = m_params.currentGroup->find(keyName);
```
Wohlstand, line 683: Don't miss the case insensitivity support:
https://github.com/Wohlstand/TheXTech/blob/bd7de09274a49d221055a2edc3877934a5310712/lib/IniProcessor/ini_processing.h#L95-L99

lib/IniProcessor/ini_processing.cpp
ds-sloth: Thanks for the notes; for case sensitivity, I wondered about that, but `hasKey` doesn't have it either. Should we add it there too?

Wohlstand: Ye, it should bein added too. That's a piece that I mised up by myself :eyes:

0lhi:
It would be best to have 1.3.6.1 add the new flag *without* removing the old one:

* Pre-1.3.6 would only read the old flag.

* 1.3.6 and 1.3.6.1 prioritize the new flag if both are included.

* 1.3.7+ will only read the new flag.

This way compat.ini files can be safely updated without breaking compatibility with older versions.

ds-sloth:
I did that. This also lets us make "Yes" the default option since it is now non-destructive.

0lhi:
I also think there shouldn't be any changes for thextech.ini, as the file is already pretty cluttered, and flags for this autoconversion would lose any meaning at 1.3.7, potentially leading to confusion.

The "Always" and "Never" options should just apply for the remainder of a session.

ds-sloth:
That's a good point. The screen is pretty quick to get through if needed.

That said, I think applying for the rest of the session might confuse the users; perhaps we should just remove the options altogether? It's already the case that the user will never be asked twice in a session for a single file.

0lhi:
Yeah I agree, removing is fine. At least for "*Always*". Maybe "*Never*" could be renamed to "*Ignore this Session*".

ds-sloth:
"Ignore this session" is a bit long, and I also think that "Always" is probably a more common use case than "Never".

The amount of these files in the wild is tiny; I think the only way a user might get the same message multiple times in a session would be if there was a compat.ini file that had been copy-pasted for all the levels in an episode.

If we're not going to include a preference (fine by me) then I think it would probably be better to just stick with "Yes" / "No".

ds-sloth:
Ah, now I understand. Anyway, this should make the whole experience a bit cleaner which I am always in favor of.

0lhi (approved):
Good for Merge 👍

One thing I noticed is that when a read-only compat.ini includes multiple deprecated flags, the user gets told about the unwritable-message multiple times. Since this measure is temporary I think it's fine.

Wohlstand (approved):
I reviewed the stuff. Right now I don't have any objections to merge this. I would like to check this more detailed later. And let other testers to verify its work on their hardware, so, they could report some.

ds-sloth:
Thank you for your review, I'll merge now.

> I would like to check this more detailed later.

I'll open an issue for this so we can remember to do so before the 1.3.6.1 release.

PR #511, opened 2023-04-08
StdPicture refactor

ds-sloth:
- Destructor of StdPicture now ensures that pictures are unloaded (reduces possible leaks)
- Made it legal to copy the public + load portions of StdPicture (used a base class for these), illegal to copy the StdPictureData portion (reduces possible use-after-free)
- Totally removed the separate loadPicture_1x functions, instead use an optional scaleFactor parameter which is used to multiply the logical w / h members
- Enable the TtfFont class to reload glyphs if their textures have been lost

With all of this, it is now possible to run clearAllTextures() in the middle of the game, and all game state will survive. This will enable hot-swapping between renderers or recovering a lost renderer in future versions.

This is an entirely technical PR and should have no user-facing changes, except for the future features it enables.

### DIFF
--- a/src/core/sdl/render_sdl.cpp
+++ b/src/core/sdl/render_sdl.cpp
@@ -91,6 +91,7 @@ bool RenderSDL::initRender(const CmdLineSetup_t &setup, SDL_Window *window)
         pLogWarning("Failed to initialize V-Synced renderer, trying to create accelerated renderer...");
 
         // fallthrough
+    default:
     case RENDER_ACCELERATED:
         renderFlags = SDL_RENDERER_ACCELERATED;
         g_videoSettings.renderModeObtained = RENDER_ACCELERATED;
@@ -378,9 +379,6 @@ void RenderSDL::loadTexture(StdPicture &target, uint32_t width, uint32_t height,
     SDL_Surface *surface;
     SDL_Texture *texture = nullptr;
 
-    target.d.nOfColors = GL_RGBA;
-    target.d.format = GL_BGRA;
-
     surface = SDL_CreateRGBSurfaceFrom(RGBApixels,
                                        static_cast<int>(width),
                                        static_cast<int>(height),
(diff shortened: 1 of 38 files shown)


src/core/sdl/render_sdl.cpp
```diff
@@ -398,13 +396,20 @@ void RenderSDL::loadTexture(StdPicture &target, uint32_t width, uint32_t height,
...
-    m_textureBank.insert(texture);
+
+    target.d.nOfColors = GL_RGBA;
+    target.d.format = GL_BGRA;
+
+    target.d.w_scale = static_cast<float>(width) / target.w;
+    target.d.h_scale = static_cast<float>(height) / target.h;
+
+    m_loadedPictures.insert(&target);
+    D_pLogDebug("RenderSDL: loading texture at %p, new texture count %d...", &tx, (int)m_loadedPictures.size());
```
Wohlstand: There is build failure because of `tx` thing that caome from nowhere :eyes:

Снимок экрана от 2023-04-17 01-14-49

Wohlstand (commented):
There is one build failure I found, I quickly fixed it

src/core/sdl/render_sdl.cpp
ds-sloth: Oops, I must have converted it to `D_` before refactoring the code a final time. Thanks for the fix :+1:

Wohlstand: The `D_` means works when it's a debug build, so, it's fine when it's printed at debug-only builds.

src/fontman/ttf_font.cpp
```diff
@@ -93,10 +93,6 @@ TtfFont::TtfFont() : BaseFontEngine()
 
 TtfFont::~TtfFont()
 {
-    for(StdPicture &t : m_texturesBank)
-        XRender::deleteTexture(t);
-    m_texturesBank.clear();
```
Wohlstand, line 98: How loaded TTF glypth textures now gets erased now?

ds-sloth: As long as `VPtrList`'s destructor correctly invokes all members' destructors, the pictures will all be unloaded via their destructors.

Wohlstand (approved):
Okay, I reviewed the whole stuff, and seems it's fine to merge now.

PR #539, opened 2023-06-10
Initial version of OpenGL renderer

ds-sloth:
image

Basically, this is a simple OpenGL renderer that scales to all versions of OpenGL from 1.1 to 4.x.

The OpenGL renderer may be enabled by (1) `config.ini`, (2) the command line, or (3) the in-game options menu. It will not be enabled by default until version 1.3.7.

It does not currently add any features to the game except proper support for SMBX 1.3-style bitmask rendering (see above). In the future (1.3.7) it will be used to add support for custom user shaders, particle systems, transition effects, etc.

It improves performance vs SDL2 on some systems, and decreases it on others. Any performance test results would be very helpful.

It should be fully ready to merge, but we should test for compatibility across multiple systems and fix any issues that arise.


.github/workflows/android-ci.yml
```diff
@@ -9,6 +9,7 @@ on:
   pull_request:
     branches:
       - main
+  workflow_dispatch:
```
Wohlstand, line 12: Has this line any sense? Or you left it by accident after some local experiments?

ds-sloth: Not important anymore, but it's very useful if you want to trigger a CI run for a branch without an open PR. We could add it to all workflows if you like, or just get rid of it here.

ds-sloth:
All CI modules fixed. OpenGL functionality verified on all supported targets -- Linux 32-bit / 64-bit / arm64; macOS x86_64; Windows 32-bit (XP) / 64-bit (10/11) / arm64 (11); Emscripten (Linux / iOS 15 Safari); Nintendo Switch.

Left to do before review/merge ready:
- Add in-game selection for preferred OpenGL version (in main settings screen for now)
- Fix Windows glitch in which OpenGL renderer context is lost during toggle render resulting in infinite `glGetError` loop

Wohlstand:
One thing I forgot to test - the Mac Mini PPC to see how OpenGL works here.

ChristianSilvermoon:
Random testing on my side

The `framerate` code does nothing.
The game can still be launched via `thextech -m -r opengl` and the frame counter will show, however

`togglerender` does not play sound-6 like other codes

ds-sloth:
I just pushed some in-progress work to add a menu option for the renderer selection.

It seems that the Windows glitch has to do with the fact that re-using windows to create different render contexts is unsupported in Windows. I might consider trying to recreate the window when changing the renderer.

ds-sloth:
Window recreation works on some (but not all) platforms:
* It works fine on Windows but is not 100% stable (repeatedly switching renderers resulted in a few segfaults for me).
* It seems to not work properly on Emscripten (and it didn't work previously, when the renderer but not the window was switched).
* Works fine on Linux (I guess this is where the SDL devs test the most frequently, so unsurprising)

Haven't tested on Android yet.

Assuming there isn't an easy fix for the instability / failures, one option would be to check which platforms support hotswapping the render/window system, and put a messagebox on the other platforms saying "Changes will not take effect until next restart."

ds-sloth:
Update: I've tested and confirmed the status of hotswapping the renderer / window provider on the following platforms.

* Linux: safe
* macOS: safe
* Android: safe
* Windows: works, unstable
* Emscripten: does not work at all

A large proportion of our user base is on Windows, so it's unfortunate that SDL can't switch between D3D9 and OpenGL contexts more stably. Should I try to work more to understand what's causing hotswapping to be unstable there?

ds-sloth:
I've now disabled renderer hotswapping except on supported systems, with a friendly "Restart engine for changes to take effect." message.
I also added back a `togglemask` cheat for testing how bitmasks look. I consider this fully ready for merge, but I would appreciate tests and reviews from @0lhi and @Wohlstand before we do so.

Wohlstand:
I'll try to run this on my Mac Mini G4 again, just to ensure the work had been fixed.

0lhi:
Scr_2023-08-11_18-24-53

1. The Dev Build notes at the top should be only visible during the Main Menu, Screenshots and Recordings.
2. Below "View Credits" the Player should be informed that this is a Beta Feature.
3. `togglemask` cheat should inform player of purpose when typed in HW/SW Mode.
4. SFX should be something shorter like "pswitch.ogg"

ds-sloth:
> (1) The Dev Build notes at the top should be only visible during the Main Menu, Screenshots and Recordings.

Good idea, I'll shorten this info now and restrict it to the menu. (Might move it to the debug screen after it goes out of beta.) I don't think there's any mechanism to render specifically to screenshots and recordings, and I don't see a big reason to add this info to them.

> (2) Below "View Credits" the Player should be informed that this is a Beta Feature.

Sure, but I don't want to need to translate a string just saying that. Is the word "Beta" universally understood? I'll add it to the info string at the top of the screen.

> (3) togglemask cheat should inform player of purpose when typed in HW/SW Mode.

Good idea, done. I think the fact it's in English should be fine, but again, I don't want to need to localize the little details here, so if it's a problem, I might rather just play the SFX_Block effect if an unsupported renderer tries to use the cheat.

> (4) SFX should be something shorter like "pswitch.ogg"

I wanted it to be different for enable and disable (to clarify in case no bitmasked graphics are onscreen), but if you think this is definitely an improvement I can do it.

0lhi:
Scr_2023-08-20_18-02-45

Cheat needs to be disabled when leaving an episode.

0lhi:
@Wohlstand To me it looks ready to merge.

Wohlstand:
Going to review... NOW.

Wohlstand (approved):
Okay, in general, this looks fine. I better to check this more after the merge.

PR #543, opened 2023-06-15
Added an ability to load custom fonts at episode or data directory (at "fonts" sub-directories)

### LINKED ISSUE #526: Allow custom fonts for new font engine at episodes/levels
Since the new font engine system had been added, there is an ability to customize fonts (or even add new fonts) would be needed.

Wohlstand:
This update adds an ability to load custom fonts for the new font engine, placed at the episode or at level/world data directories. Additionally, it tracks what current fonts set is loaded, and if needed, it does reloading of requested fonts once the font path becomes different.

Fixes #526

### DIFF
--- a/src/fontman/font_manager.cpp
+++ b/src/fontman/font_manager.cpp
@@ -31,12 +31,14 @@
 #include <Graphics/graphics_funcs.h>
 #include <IniProcessor/ini_processing.h>
 #include <Logger/logger.h>
-#include "core/render.h"
-
 #include <Utils/files.h>
 #include <DirManager/dirman.h>
 #include <IniProcessor/ini_processing.h>
 #include <fmt_format_ne.h>
+
+#include "global_constants.h"
+#include "core/render.h"
+
 
 #include <vector>
 #ifdef LOW_MEM
@@ -82,9 +84,9 @@ static TtfFont         *g_defaultTtfFont = nullptr;
 static bool             g_fontManagerIsInit = false;
 
 #ifdef LOW_MEM
-typedef std::map<std::string, int> FontsHash;
+typedef std::map<std::string, vbint_t> FontsHash;
 #else
-typedef std::unordered_map<std::string, int> FontsHash;
+typedef std::unordered_map<std::string, vbint_t> FontsHash;
 #endif
 //! Database of available fonts
 static FontsHash        g_fontNameToId;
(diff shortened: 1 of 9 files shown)


src/fontman/font_manager.cpp
```diff
@@ -63,19 +79,101 @@ static bool             g_fontManagerIsInit = false;
 typedef std::unordered_map<std::string, int> FontsHash;
```
ds-sloth, line 79: This is an unexpectedly expensive data structure (might easily be 1 KB each even with a small number of fonts loaded).

src/fontman/font_manager.cpp
```diff
@@ -63,19 +79,101 @@ static bool             g_fontManagerIsInit = false;
...
 //! Database of available fonts
 static FontsHash        g_fontNameToId;
+//! Backup of the database of available fonts
+static FontsHash        g_fontNameToIdBackup;
+static FontsHash        g_fontNameToIdBackupWorld;
 
 static bool             g_double_pixled = false;
 
+static const int c_smbxFontsMapMax = 100;
```
ds-sloth, line 88: It's a good idea to make a static limit, but I think this should depend on whether scripting is enabled (4 like SMBX if not enabled, 8-16 if enabled).

src/fontman/font_manager.cpp
```diff
@@ -122,66 +220,46 @@ TtfFont* FontManager::getTtfFontByName(const std::string& fontName)
 #endif
 }
 
-void FontManager::initBasic()
-{
 #ifdef THEXTECH_ENABLE_TTF_SUPPORT
-    if(!g_ft)
-        initializeFreeType();
-    g_defaultTtfFont = nullptr;
+static bool s_loadFintsFromDir(const std::string &fonts_root,
```
ds-sloth, line 224: Advice: it's better to take an already-loaded DirListCI (`g_dirEpisode` or `g_dirCustom`, or a temporary one in the global case) as an argument to this, to avoid the expense of relisting a directory (high on some of our platforms).

ds-sloth (changes requested):
I've read everything. Here are my main notes / requests. After merging, I'd like to verify the resulting resource usage, but any other optimizations can happen after merge.

src/fontman/font_manager.cpp
ds-sloth: Since we currently have no Lua scripting, we can make it universally 4 until v1.3.7.

Wohlstand: 16 better, 4 is bad, because it kills the ability to add new fonts without replacing of existing (custom fonts were already can be used at the LunaScript in a condition the "font-override" entry had been added to assign the font ID).

src/fontman/font_manager.cpp
Wohlstand: Good idea, I'll try this.

src/fontman/font_manager.cpp
Wohlstand: Now the question: what is more memory-effective? The hash-table (unordered_map) or a red-black tree (map)?

Wohlstand: ... trying to find something like: https://stackoverflow.com/questions/9375450/stdunordered-map-very-high-memory-usage

src/fontman/font_manager.cpp
LoveBodhi: I tried Font 5, works as well... (Message box untested)

LoveBodhi: Should 8 fonts fine for low-end device?

src/fontman/font_manager.cpp
Wohlstand: I made std::map being used on low-mem ends, should be better. Right?

src/fontman/font_manager.cpp
Wohlstand: Okay, I tweaked this thing, and also I replaced `int` below with the `vbint_t` that uses int16_t type on low-end devices.

src/fontman/font_manager.cpp
Wohlstand: Anyway, here is a dynamical search of files in a directory by the ".font.ini" template to find out raster fonts, and the, finding TTF fonts from a manually-filled list in the INI file.

Wohlstand: @ds-sloth, what do you think about this?

Wohlstand: Okay, I updated the thing:
- I extended the DirListCI to support necessary cals
- I reworked the font manager side to use global DirListCI instances
Should work now!

Wohlstand:
@ds-sloth, just now I completed all your notices, so, what do you think about the thing now?

ds-sloth (approved):
I've checked all of the changes, the design is clean, I am happy for it to be merged.

Wohlstand:
Super!
Going to merge now.

PR #550, opened 2023-07-10
Enable keyhole fix permanently + gameplay timer initialisation fix

Wohlstand:
For the future speedrun needs, there are next changes done:
- The keyhole framerate fix became a permanent update with no way to disable (The compatibility flag has been removed).
- Changed the gameplay timer initialisation: begin the counting of cycles with 0 rather than with 1.

@0lhi, please test the whole thing and give a review on the whole thing.


0lhi:
I made a new Unit Test including two variations of KeyHoleTest - the difference between the two being that the first and second levels are switched: [KeyHoleTest.zip](https://github.com/Wohlstand/TheXTech/files/12001731/KeyHoleTest.zip)

Here's the situation on TXT 1.3.6:

Insta-End Level: *00.015 / 00.000* ❌

Scr_2023-07-10_14-12-58

Run-Level: *01.965 / 01.965* ✔️

Scr_2023-07-10_14-13-28

Here's the situation in the current PR:

Insta-End Level: *00.000 / 00.000* ✔️

Scr_2023-07-10_14-23-29

Run-Level: *01.950 / 01.965* ❌

Scr_2023-07-10_14-40-14

We have three goals to accomplish:

1. The Level Timer must always report the same time for auto-runs.
2. Both variations of KeyHoleTest must report the same Final Time (either 11.731 or 11.746)
3. Comparisons against SMBX-R must remain Valid.

0lhi:
@ds-sloth When restarting inside a test, the game is searching for controllers again, ignoring currently pressed buttons. This is a run-killer. The game must consider inputs from the very first frame.

@Wohlstand On that matter, restarts are too fast. Chances are the level already began before the Player can take the perfect starting position (in case of the KeyholeRun, Run+Right). There should be an artificial loading time of at least one second.

Wohlstand:
About the second test, why it counted less than before: just because at 1.3.6, all timers had to start from 1. Now all timers start from 0. The 1 was a workaround in order to match the buggy timer for SMBX2 that counted an extra frame.

0lhi:
> The 1 was a workaround in order to match the buggy timer for SMBX2 that counted an extra frame.

Yeah, so, we don't care about that anymore. There has to be a different reason why it has a different behavior on a new game though, no matter if it's 0 or 1? Do you have an idea what it could be?

ds-sloth:
Added a number of changes (described in their commit messages) to try to address @0lhi's comments here.

Note that these changes are touching a large number of game modules, so I'd prefer us to merge this branch via the **Rebase and merge** option so that all of the commits show up on the main tree, instead of one big and hard-to-blame commit. This also means it would be nice to avoid doing any `main` -> `wip-keyhole-permafix` merges if possible.

ds-sloth:
@0lhi can you please explain the case you marked with an :x: in your recent message? How would that be fixed?

ds-sloth:
@0lhi, thanks for explaining the situation on Matrix. The problem was that the episode-first frame was treated differently from other frames in the gameplay timer, and I fixed that.

At your convenience can you please run your tests again and verify that the logic is fixed?

0lhi:
Logic is perfect now 👍

UI is still hard on the eyes though.

Scr_2023-07-23_14-13-11

What's the information below the framerate? Is that needed here?

Can you make it so that "P1 Disconnected" and the Input Display are never simultaneously onscreen?

Can you make it so that the Character is displayed in a better visible position?

Scr_2023-07-23_14-14-09

Can you make it so that the "P1 Connected" message never makes it to the next level? Ideally the very first run starts without it.

Also, the Level Timer shouldn't increase post-level. Episode Timer probably isn't needed at all.

ds-sloth:
> What's the information below the framerate? Is that needed here?

Frame time as a percentage of target time, average and max. The whole performance profiling system could use a graphical overhaul before the 1.3.7 release. These numbers were added to the main screen (instead of the debug info screen) because the debug info screen covers the entire screen at low resolutions. For 1.3.6.1, I'll just disable these frame time measures. I'll make a separate issue about this.

> Can you make it so that "P1 Disconnected" and the Input Display are never simultaneously onscreen?

Sure. That'll just require adding a check to the input display for whether the quick reconnect screen is active for that player.

> Can you make it so that the Character is displayed in a better visible position?

This is a glitch from accidentally rendering a non-existent "level"; the background is meant to be black. I'll try to fix that. If we want to display the character, we should do it intentionally. (Takes a bit of work.)

> Can you make it so that the "P1 Connected" message never makes it to the next level?

Good idea. Easy enough to do it for this special case.

> Also, the Level Timer shouldn't increase post-level.

Great note. Easy to do using the negative `LevelBeatCode` values I use to signal a start / restart testing menu.

> Episode Timer probably isn't needed at all.

Good point. I'll just disable the display of it during level tests.

ds-sloth:
@Wohlstand, @0lhi asked me to add outlines to the text on the testing menu for readability. I did that by backporting and polishing some code from the config branch. SMBX Font ID 5 is now a copy of font 4 with outlines added. Please review the most recent commits and make sure that they seem good to you.

ds-sloth:
@0lhi, I think Matrix is down. I just fixed that timer flashing issue. Are we good to merge from your perspective?

0lhi:
There's still the issue that returning to the Menu counts an Extra Cycle if no Exit is reached.

Scr_2023-08-01_00-08-14

This is mildly irritating and displays a slightly inaccurate end-time for levels that canonically end with a death.

---
Also found another, more serious thing:

Scr_2023-08-01_00-19-32

Currently, levels with entrance warps but no start position cannot be tested. Example shows warp5.lvlx from Invasion 2.

X2 provides a solution for this:

ds-sloth:
@Wohlstand, @0lhi please review the most recent commit, which starts the player death fadeout effect before the level is about to end.

Advantages:

(1) Keeps gameplay duration identical to SMBX64 duration
(2) Doesn't pause speedrun timer
(3) Doesn't pause moving NPCs
(4) Removes nested frame loop
(5) Same code structure as warp fades

0lhi (approved):
> We could consider calling it at the beginning of the frame instead. What do you think?

What advantages would that bring? In PM I was under the impression that this was needed to resolve the death's extra-cycle, but it appears not?

ds-sloth:
> What advantages would that bring?

It's mostly juts a way to keep it very organized and controlled. It means that we measure every frame from the start, and we measure a run's length by the number of frames before victory conditions are achieved. That's a bit more organized than measuring every frame from the moment drawing occurs.

ds-sloth:
We discussed this point further internally and agreed to keep the tick immediately before the draw call for consistency with previous TheXTech versions and SMBX2.

I'm happy with everything here and ready to merge as well.

ds-sloth (approved):

src/fontman/font_manager.cpp
```diff
@@ -575,39 +575,53 @@ void FontManager::printText(const char* text, size_t text_size,
                             int x, int y,
                             int font,
                             float Red, float Green, float Blue, float Alpha,
-                            uint32_t ttf_FontSize)
+                            uint32_t ttf_FontSize, bool outline)
```
Wohlstand, line 578: Maybe you'll also provide an option to specify a colour?

src/graphics/gfx_print.cpp
```diff
@@ -98,60 +101,106 @@ int SuperTextPixLen(int SuperN, const char* SuperChars, int Font)
...
+    bool outline = false;
+
+    if(Font == 5)
+    {
+        Font = 4;
+        outline = true;
+    }
+
+    if(Font == 5)
+        Font = 4;
```
Wohlstand, line 162: Are you reserving the font 5 for internal needs?

src/graphics/gfx_update.cpp
```diff
@@ -2275,6 +2271,9 @@ void UpdateGraphics(bool skipRepaint)
                 XRender::offsetViewportIgnore(true);
                 SuperPrint(fmt::format_ne("{0}", int(PrintFPS)), 1, 8, 8, 0.f, 1.f, 0.f);
 
+#if 0
+                // deprecated profiling info
+
```
Wohlstand, line 2276: You better put the comment at the `#if 0` itself, like:
```cpp
#if 0 // deprecated profiling info
```

src/main/screen_connect.cpp
```diff
@@ -109,6 +109,7 @@ void Reconnect_Start()
         s_playerState[i] = PlayerState::Disconnected;
         s_menuItem[i] = 0;
         s_inputReady[i] = false;
+        g_charSelect[i] = Player[i+1].Character;
```
Wohlstand, line 112: Please add spaces between operators: `i + 1`.

ds-sloth: Good note, done it everywhere.

src/graphics/gfx_update.cpp
ds-sloth: Good note, done.

src/graphics/gfx_print.cpp
ds-sloth: Yes, that's right. Let me see... do you know if there is a way to check whether there is a user-specified Font 5? It would be nice to make this hardcoded re-use only occur if the default Font 5 would be used.

src/fontman/font_manager.cpp
ds-sloth: Great idea, done.

src/main/menu_controls.cpp
```diff
@@ -1068,19 +1068,19 @@ int menuControls_Mouse_Render(bool mouse, bool render)
                     SuperPrint(name, 3, sX+32+(width-32)/4, start_y+(b_base+2*i - scroll_start)*line);
 
                     if(MenuCursor == i && !s_secondaryInput && g_pollingInput)
-                        SuperPrint("...", 3, sX+32, start_y+(b_base+2*i+1 - scroll_start)*line);
+                        SuperPrint("...", 3, sX+32, start_y+(b_base+2*i + 1 - scroll_start)*line);
```
Wohlstand, line 1071: `b_base + (2 * i) + 1`

Wohlstand (commented):
Looks good now, if not count one minor style thing. I can't approve myself as I created this PR by myself :P

src/main/menu_controls.cpp
ds-sloth: Did a full refactor of this section, thanks a lot for this note.

PR #561, opened 2023-07-18
Language Switching UI

ds-sloth:
Doing this as a PR instead of directly to main because
(1) it slightly changes the order of attempts to load the language, and
(2) it officially requires translations to have the form `thextech_[lang]-[locale].json` or `thextech_[lang].json` (lowercase, without any underscores).

Please confirm that both of those are okay with you.

@Wohlstand, please review at your convenience. It doesn't touch much other code, so there shouldn't be a rush.

One note: some of the languages currently included in the repository are very incomplete. It would probably be useful to only include languages with general completeness in the 1.3.6.1 release.

Scr_2023-07-18_16-23-25


Wohlstand:
About languages: these are blank or has a very few number of translated strings, I'll don't include into public assets. Its possible to heuristically detect which translations are blank or incomplete by relative file size check compared to English.

Wohlstand:
On a quick test, looks like a good thing. However, I would like to add the skipping any JSON files that has file size less than specific bytes number (blank translations do have 3 bytes because they contain only the `{}` as WebLate doesn't includes blank lines into exported JSONs).

ds-sloth:
Just did this with a limit of 8 bytes, also pre-loaded the translation locations during game load (should speed up the switching process, but costs ~1 KB RAM).

src/core/language/language_common.cpp
```diff
@@ -88,6 +92,87 @@ static bool detectSetup()
...
 
 
+void XLanguage::findLanguages()
+{
+    constexpr size_t prefix_length = 9; // std::string("thextech_").size();
+
+    // clear and free s_languages
+    {
+        std::vector<std::string> blank;
+        std::swap(s_languages, blank);
```
Wohlstand, line 102: What the problem with the `.clear()` ?

Wohlstand: I see (https://stackoverflow.com/questions/35514909/how-to-clear-vector-in-c-from-memory).
I guess, you would make a generic util call that can be used at other places too.

ds-sloth: Doesn't free the RAM, could be important if the number of languages shrinks significantly between calls. But in the ordinary cases, `clear()` should be fine. We could replace it for readability.

PR #564, opened 2023-07-24
Separated SDL Init/Quit + Timestaps per logger event

Wohlstand:
There are two changes made:
- SDL init/quit has been moved into separated thing: SDL is much more than just a window, so, it should be initalized globally before everything, and should be closed after everything got been closed too.
- Added timestamps to every log entry written into the file: the time will show how much time was passed since application startup. This could help to debug the performance at certain points.

Note: if SDL is completely not needed, the `THEXTECH_NO_SDL_CORE` macro should been defined, and the `sdl_core.cpp` should not be included into the build.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -672,6 +672,7 @@ if(NINTENDO_3DS)
         src/core/3ds/msgbox_3ds.cpp
         src/core/3ds/events_3ds.cpp
 #        src/core/3ds/3ds-audio-lib.cpp
+        src/core/sdl/sdl_core.cpp
     )
 elseif(NINTENDO_WII)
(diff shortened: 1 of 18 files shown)


ds-sloth:
Looks good to me, make sure that SDL isn't initialized twice at targets that use MixerX but not SDL window.

I did something similar (separated SDL window init/destruction from SDL init/destruction) at the OpenGL branch, I'll update that after we merge here.

Wohlstand:
Made small update to ensure nothing unexpected gets initialized on Wii and 3DS

ds-sloth:
Cool, should be fine to merge now.

ds-sloth (approved):

PR #570, opened 2023-07-30
Experimental CI workflow for asset conversion

ds-sloth:
(Ignore for now, for testing purposes only...)


ds-sloth:
I don't need all the other workflows to trigger, let me delete this PR and set the workflow to activate for this specific branch...

PR #571, opened 2023-07-30
CI for the homebrew assets

ds-sloth:
Looks like it's working! Please review and check that everything looks fine.


Wohlstand:
Question: did you performed translations update?
I mean, do:
```bash
ASSETS_ROOT1="...path to assets root..."
cd .github/ci-helper
bash translate_update.sh "${ASSETS_ROOT}"
cd ../..
```
That script puts updated translations into assets, and applies patches are supplied at mainstream assets.

Wohlstand:
At my desktop builds you may found the `name: Apply update to translations` step, that actually updates translations and applies patches (character names, etc.)

ds-sloth:
I just added that now, thanks for the note!

.github/workflows/homebrew-assets-ci.yml
```diff
@@ -0,0 +1,189 @@
...
+
+jobs:
+  build:
+    name: ${{ matrix.source.name }} for ${{ matrix.target.name }}
+    runs-on: ubuntu-latest
+    container: ${{ matrix.target.docker_image }}
+    env:
+        DEVKITPRO: /opt/devkitpro
+    strategy:
+      fail-fast: false
```
Wohlstand, line 26: If this set to `false`, that means, if any job will fail, the chain will continue. That means, thing would turn into the mess if one job depends on a previous.

Issue #122, opened 2021-05-28
Feature Request: Dragon Coins on HUD


ChristianSilvermoon:
There's no way to know how many Dragon Coins you've collected in a stage other than remembering them. It would be nice if they displayed on the HUD somewhere like in Super Mario World, the graphic used in SMW is already present on the HUD for the coin counter.

Wohlstand:
Good idea, at least, make that work in next conditions:
- the compat.ini option enables it
- all necessary resources were presented
- add an option that will make them don't appear again after collection
Also, I MUST to make dragon coins to NOT transform on a P-Switch

Wohlstand:
It's needs to be made carefully as dragon coins use can be abused in some cases.

ChristianSilvermoon:
The only reason I can imagine for not making them appear on the HUD in certain episodes is that it might break immersion slightly if they've been re-sprited, but realistically it shouldn't impact gameplay itself at all if added :thinking:

LoveBodhi:
Remind me "Star Coins"!

P-Star7:
HUD asset exists already (SMW coin logo), but users may want a different graphic, and also maybe "uncollected" logo to count all the remaining coins. I wonder if this option should be expanded to other NPC's deaths, since Dragon Coin's death function is what will be read for the HUD change anyways.

ChristianSilvermoon:
> HUD asset exists already (SMW coin logo), but users may want a different graphic, and also maybe "uncollected" logo to count all the remaining coins. I wonder if this option should be expanded to other NPC's deaths, since Dragon Coin's death function is what will be read for the HUD change anyways.

If going that direction, I think it'd be better to support adding/manipulating HUD elements through events rather than making it a part of Dragon Coin tracking, which be just showing info on screen that the game already tracks.

ds-sloth:
Has been implemented using the score counter for dragon coins. Maximum of 8 (5 UP) shown on screen (at which point they award 5 UP).

I think that we should certainly use the SMW coin logo because (1) that is how SMW does it, and except in rare cases this game is inspired by the innovations of that series; (2) assets are already designed to make SMW coin logo visible against all background (since it is used for coins); and (3) that minimizes the number of assets that must be loaded at one time.

Question: what should the positioning be? Ideally, the coin and life counters should have the same height, and ideally the HUD would not expand too much. This is the best I can think of right now, but there may be other better approaches.

[...]

PR #579, opened 2023-08-06
Medal tracking

ds-sloth:
This PR allows TheXTech to track acquired medals (requested a long time ago in #122).

This includes:
(1) A new save structure that keeps track of how many stars and medals are in each level, and how many medals have been collected. (Has been replaced by @Wohlstand with a permanent save structure.)
(2) Added a new config option `"medals-show-policy": ["hide", "collected-only", "counts-only", "show-full"]`. Available and collected medals can be displayed in the upper-right corner of the level HUD, and above each warp / level (similar to stars).
(3) If `"show-full"` is chosen, then the order specified by the medals' `special_data` attribute (eventually settable in the editor) will be used to order the medals. If the attribute is unspecified, then medals will be ordered automatically based on their distance from the level entrance.

Uses a new asset, `graphics/ui/Medals.png`: Medals  Medals

I'm opening this now to get public feedback and comments on the feature' design.

The main question here is whether the "Shiny" medals from the current version should exist. A medal is shiny if you got it without dying, and the level only tracks the shiny medals from your single best run (you can't combine shiny medals from multiple playthroughs of a level). Does that idea make sense to people? Is there some other "best" tier of medal we should consider?


LoveBodhi:
Will 10-Coin (SMB Wonder) be doable in the future with this?
10 PurpleCoin HUD

ds-sloth:
> Will 10-Coin be doable?

This PR is for tracking the medals (collectible coins) that the player has gotten.

If 10-coins will be collectible in your asset pack, then a future scripting functionality could be used to mark 10-coins as a collectible item to be tracked, but that isn't included in this current PR.

src/main/level_save_info.cpp
```diff
@@ -0,0 +1,255 @@
...
+}
+
+static bool s_parseSaveInfo(LevelSaveInfo_t& info, const std::string& s)
+{
+    info = LevelSaveInfo_t();
+
+    unsigned version = 0;
+    int end_pos = 0;
+
+    if(sscanf(s.c_str(), "V:%2u;%n", &version, &end_pos) != 1 || end_pos < 1)
```
Wohlstand, line 73: Please use `SDL_sscanf`

ds-sloth: Ah, good note, thanks. I generally forget which functions have SDL versions. But, I'll delete this whole file when I update it to use PGE-FL.

Wohlstand: Little note: recently I already added new stuff into PGE-FL, please take a look.

ds-sloth:
Should be ready for review.

The main question here is whether the "Shiny" medals from the current version should exist. A medal is shiny if you got it without dying, and the level only tracks the shiny medals from your single best run (you can't combine shiny medals from multiple playthroughs of a level). Does that idea make sense to people? Is there some other "best" tier of medal we should consider?

ds-sloth:
I've done a full pass over the full diff and polished everything. I found a few minor bugs and fixed them as well. This is fully ready for review and merge on my side.

Wohlstand (commented):
In general looks good, but there are next things needed to be added:
- A config, add the "undefined" value (-1), that should being default
- At gameinfo, add the same field, that will be OFF by default.
- At compat.ini, add the same field, and it should have "undefined" value by default.

Treat by next order:
- **game info** reads first. It defines the base default value for assets.
- **compat.ini** reads next. If not undefined, it overrides. It defines the base default for episode.
- Read from config. If not undefined, it overrides everything.

ds-sloth:
Good note, I agree that should be needed before final release.

However, since this won't be released until 1.3.7, let's wait for the Config Update to handle that instead of spending more time making config merge logic that will be replaced during the Config Update.

Wohlstand:
Then, I think, should be fine: feel free to put the config branch as a draft for a reviewing/polishing, and make it being normal PR once you feel it's ready.

LoveBodhi:
Should this can be different sizes (such as 32x32 medal GFX)?

ds-sloth:
Note: I have revised the default medals asset from Medals Medals for the SMBX64-based asset pack, and added a new medals asset for the AoD asset pack.

Wohlstand:
One note: after this thing gets merged, updated the Wiki please to document this feature. Just to don't forget to update this in the future release.

src/globals.h
```diff
@@ -1715,6 +1755,12 @@ extern bool ExitMenu;
 //Public LevelSelect As Boolean 'true if game should load the world map
 extern bool LevelSelect;
 
+//! NEW: utility function to check if in a hub level
+inline bool InHub()
```
Wohlstand, line 1759: Suggested to use the `SDL_FORCE_INLINE` thing here.

src/graphics/gfx_hud.cpp
```diff
@@ -412,6 +426,159 @@ void DrawInterface(int Z, int numScreens)
...
+    int coin_height = GFX.Interface[2].h;
+
+    if(GFX.Medals.inited)
+    {
+        coin_width = GFX.Medals.w / 4;
+        coin_height = GFX.Medals.h;
+    }
+
+    if(max > 8)
+        max = 8;
```
Wohlstand, line 500: Maybe define this constant at `global_consts.h`?

src/main/level_medals.cpp
```diff
@@ -0,0 +1,282 @@
...
+    return nullptr;
+}
+
+void CurLevelMedals_t::get(uint8_t idx)
+{
+    // don't save star coins in hub level
+    if(InHub())
+        return;
+
+    if(idx < 8 && idx < max)
```
Wohlstand, line 79: Should 8 become a constant at global constants header?

src/main/level_medals.cpp
```diff
@@ -0,0 +1,282 @@
...
+
+    if(!info)
+        return;
+
+    info->medals_got |= got;
+
+    int old_best_count = 0;
+    int new_best_count = 0;
+
+    for(int i = 0; i < 8 && i < info->max_medals; ++i)
```
Wohlstand, line 163: Here too, 8 should be a kind of `c_max_medals` global constant.

src/main/level_medals.cpp
```diff
@@ -0,0 +1,282 @@
...
+        std::sort(auto_medals.begin(), auto_medals.begin() + auto_medal_count);
+    }
+
+    // auto-assign indexes to unspecified medals
+    for(int auto_i = 0; auto_i < auto_medal_count; ++auto_i)
+    {
+        NPC_t& n = NPC[auto_medals[auto_i].second];
+
+        // try to find an index
+        for(int i = 0; i < 8; ++i)
```
Wohlstand, line 268: And here should be a `c_max_medals` constant.

src/main/level_medals.h
```diff
@@ -0,0 +1,78 @@
...
+    uint8_t max = 0;
+
+    //! bitfield of medals previously acquired in level (imported as "got" from save data)
+    uint8_t prev = 0;
+    //! bitfield of medals acquired during current play session (on win, combined with "got" in save data)
+    uint8_t got = 0;
+    //! bitfield of medals acquired since most recent lost life (on win, compared with and possibly replaces "best" in save data)
+    uint8_t life = 0;
+
+    //! sets specific coin as obtained
```
Wohlstand, line 51: Just a note about DoxyGen comments:
```
//! Use this format for variables only

/**
 * \brief And use this to document functions.  
 */
```
Especially because this way allows to describe arguments and allows to desctibe the sense of return value.

src/main/level_save_info.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+static bool s_exportSingleSaveInfo(saveLevelInfo& s, const LevelSaveInfo_t& info)
+{
+    if(!info.inited())
+        return false;
+
+    s.max_stars = info.max_stars;
+    s.max_medals = info.max_medals;
+
+    int size = SDL_min(info.max_medals, 8);
```
Wohlstand, line 42: Should be `c_max_medals` global constant.

src/main/level_save_info.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+    unsigned max_stars = s.max_stars;
+    unsigned max_medals = s.max_medals;
+
+    // validate limits
+    if(max_stars > 255)
+        max_stars = 255;
+
+    if(max_medals > 8)
+        max_medals = 8;
```
Wohlstand, line 68: Should be a `c_max_medals` global constant.

src/main/level_save_info.cpp
```diff
@@ -0,0 +1,234 @@
...
+static bool s_importSingleSaveInfo(LevelSaveInfo_t& info, const saveLevelInfo& s)
+{
+    info = LevelSaveInfo_t();
+
+    unsigned max_stars = s.max_stars;
+    unsigned max_medals = s.max_medals;
+
+    // validate limits
+    if(max_stars > 255)
+        max_stars = 255;
```
Wohlstand, line 65: Maybe create the `c_max_stars_per_level` constant or something like?

src/main/level_save_info.cpp
```diff
@@ -0,0 +1,234 @@
...
+        if(!used_indexes[Variant - 1])
+        {
+            medal_count++;
+            used_indexes[Variant - 1] = true;
+        }
+    }
+
+    // cap medal count at 8
+    if(medal_count > 8)
+        medal_count = 8;
```
Wohlstand, line 209: Should be `c_max_medals` constant.

Wohlstand (changes requested):
Okay, I completely reviewed, and there are notes from me:
- Please make global constant instead of magic numbers (such as maximum number of medals, and maximum number of starts per every level).
- Please try to document functions not in same way as variables: there is a possibility to describe arguments and sense of the function's return (if presented).

Other things were mostly fine, or I something missed up.

ds-sloth:
I'll update the docs and inline case soon.

For the constants, 8 and 255 are not changeable magic numbers; they're enforced by the field width used. If you like I can replace them with global constants, but it's only possible to decrease these, not increase them.

Wohlstand:
> For the constants, 8 and 255 are not changeable magic numbers; they're enforced by the field width used. If you like I can replace them with global constants, but it's only possible to decrease these, not increase them.

That's fine, just give an explanation in comments (already done) and ye, a kind of static asserts should also prevent to exceed these values.

Wohlstand (approved):
Okay, for me the thing looks fine, I'll merge it.

Just need for the future (as you said, at the config branch) to allow episode makers to define the default value that will be used everywhere if user didn't set their override.

PR #595, opened 2023-08-23
Simple pre-processor for message boxes

Wohlstand:
This is a small thing that allows to show different text depending on various conditions. Currently this implementation is able to print different branches of text depending on playable character number. I would be like to merge it quickly being very simple for now, because I want to use this logic in my The Invasion 2 translation. And later, it could be extended.

When it proceeds a message from NPC, it takes player index directly to detect player's character immediately. But when it's a message box called by event, then a heuristic logic being implemented to detect possible "source" of the event (in a case of events chain from NPC dialogue, the last player index used from NPC call, will be re-used at events too).

## Syntax
- `#if player(x)` or `#if player(x, y, z, ..., n)` - true if player matches one of comma-separated values.
- `#elif player(x)` - same as `#if`, but will gets checked if previous condition is false.
- `#else` - opens content if previous condition is false.
- `#endif` - final of conditional branch.

## The example text:
```
#if player(3)
О, Ваше Высочество! Для меня такая честь повидать Вас! У меня есть для Вас подарок, позвольте открыть сундук...
#elif player(4)
Здарова, брательник! Как поживаешь? Возьми кое-что из сундука, это тебе пригодится!
#elif player(5)
О, ты случайно не тот герой из далёкой страны, про которого все рассказывали? Возьми кое-что из этого сундука.
#else
Ты случайно не тот санте-
хник, о котором все гово-
рят? Возьми кое-что из 
этого сундука.
#endif
```

## The showcase of result:
Scr_2023-08-23_11-47-39
Scr_2023-08-23_11-46-41
Scr_2023-08-23_11-47-16
Scr_2023-08-23_11-47-59


SkullGamer205:
👀

ds-sloth:
Just added event player tracking to the main branch, and removed the local player tracking here.

Can you please check that conversation chains still work as expected in this branch?

Wohlstand:
Okay, as we talked at Discord, it's fine to merge now. I'll do that now.

PR #605, opened 2023-08-30
A few new compat.ini settings for players

SpencerEverly:
Up for review today is 3 new compat.ini settings I put together myself:

- easier-player-powerdown: Allows any player to have 2 hurt states instead of just one (3rd tier -> 2nd tier -> 1st tier). This also adds a new forced state (551) into the code as well.
- easier-player-physics: Allows better controlled physics, based off [this Lua script](https://www.smbxgame.com/forums/viewtopic.php?t=25854) (This applies to all players on an episode)
- player-faster-climbing: Allows any player to climb faster when holding run.
- enable-kind-lava: Allows lava to be easier and less avoidable, though you'll still get hurt when touching it.

This also adds a new function (mathSign), used for easier-player-physics.

Idk if this will get accepted, but oh well. Here's to another pull request by me!


Alucard648:
"easier-player-powerdown" should also give Peach, Toad and Link +2 Max Hearts for balance purposes. And also instant item box usage option would also help Mario and Luigi acceess item box contents, especially in intense boss fights

LoveBodhi:
NSMB / SMB3U Style Power-down system hmm

SpencerEverly:
I've added another compat setting as of the recent commit: "enable-kind-lava". This one should make the player get hurt rather than die when touching lava. I thought it would be a good idea to add because it might be useful for later when Lua scripting gets added, where cancelling the player harm when touching lava can lead to episode-specific things (Though it also makes lava easier and kinder in-game)

src/math.h
```diff
@@ -0,0 +1,29 @@
...
+
+#pragma once
+#ifndef MATH_H
+#define MATH_H
+
+#include <cmath>
+
+extern int mathSign(int a);
+
+#endif // MATH_H
```
Wohlstand, line 29: You can have just a template:
```
template <typename T> int mathSign(T val) 
{
    return (T(0) < val) - (val < T(0));
}
```
Take from here: https://stackoverflow.com/questions/1903954/is-there-a-standard-sign-function-signum-sgn-in-c-c

src/rand.h
```diff
@@ -103,5 +103,6 @@ inline int iRand_round(int max)
     return i/2;
 }
 
+extern int mathSign(int a);
```
Wohlstand, line 106: Absolutely redundant reference here. You made the math.h.

Wohlstand:
Please rename `math.h`, it will CONFLICT with system-wide `math.h` which is a standard C library.

src/player.cpp
```diff
@@ -7699,3 +7788,13 @@ bool SwapCharAllowed()
...
+
+bool PlayerGroundTouching(const int A)
+{
+    if(Player[A].Pinched.Bottom1 != 0 || Player[A].Slope != 0 || Player[A].StandingOnNPC != 0)
+    {
+        return true;
+    }
+    
+    return false;
+}
```
Wohlstand, line 7800: Missing new-line character.

src/player/player_update.cpp
```diff
@@ -596,18 +598,48 @@ void UpdatePlayer()
...
+                    }
                     else
-                        Player[A].Location.SpeedY = 0;
+                    {
+                        if(Player[A].Controls.Left)
+                            if(!Player[A].Controls.Run)
+                                Player[A].Location.SpeedX = -1.5;
+                            else
+                                Player[A].Location.SpeedX = -3;
+                        else if(Player[A].Controls.Right)
```
Wohlstand, line 623: You better to add braces around these blocks. Without braces, this looks like Python...

src/player/player_update.cpp
```diff
@@ -596,18 +598,48 @@ void UpdatePlayer()
...
+                            if(!Player[A].Controls.Run)
+                                Player[A].Location.SpeedX = 1.5;
+                            else
+                                Player[A].Location.SpeedX = 3;
+                        else
+                            Player[A].Location.SpeedX = 0;
+                        if(Player[A].Controls.Up && Player[A].Vine > 2)
```
Wohlstand, line 630: Needed a blank line before `if` block begin.

src/player/player_update.cpp
```diff
@@ -4590,3 +4639,42 @@ void UpdatePlayer()
...
+                            Player[A].Location.SpeedX = Player[A].Location.SpeedX + 0.08 * mod;
+                    }
+                    else if(Player[A].Controls.Left)
+                    {
+                        if(Player[A].Location.SpeedX > 0)
+                            Player[A].Location.SpeedX = Player[A].Location.SpeedX - 0.08 * mod;
+                    }
+                    else
+                        Player[A].Location.SpeedX = Player[A].Location.SpeedX * 0.98;
+                    int xSpeedDiff = Player[A].Location.SpeedX * Player[A].lastXSpeed;
```
Wohlstand, line 4669: Needed a blank line between these lines. You should add blank lines to make the code visually look better. If you don't put blank lines where they are needed, code looks like a clutter of text. So, it's better to logically separate if blocks, and various logical parts of code.

src/player/player_update.cpp
Wohlstand: Example:
```cpp
if(xxx)
{
    if(yyy)
        do();
    else
        do2();
}
else if(zzz)
{
    if(qqq)
        do3();
    else
        do4();
}
```

src/player/player_update.cpp
```diff
@@ -616,27 +619,37 @@ void UpdatePlayer()
                     else
                     {
                         if(Player[A].Controls.Left)
+                        {
                             if(!Player[A].Controls.Run)
                                 Player[A].Location.SpeedX = -1.5;
                             else
                                 Player[A].Location.SpeedX = -3;
+                        }
+
```
Wohlstand, line 628: before `else` - line not needed,
but if `if` without `else`, then line is needed.

0lhi:
Closing this PR since it tries to accomplish too many things at once. The individual points need more consideration, as there can be different preferences on how an idea should be implemented.

I'd like to encourage you to bring up your climbing and physics suggestions at #636, explaining what the changes do exactly (ideally with example GIFs), and why you think they are a worthwhile addition to the Engine.

For the HP Update, one question that I think is important to clarify is whether Chars 3-5 should be given the Item Box to make them balanced with 1-2. Other ideas such as auto-applying what's in the Item Box may also be worth exploring, which is why a separate discussion is needed before merging any change. If you like you can open one.

PR #606, opened 2023-08-31
Implemented the fallback for the case when no modern fonts provided

### LINKED ISSUE #600: Game fails to work if fonts for new font engine are missing
**Describe the issue**

Attempt to play the game with old assets without `fonts` directory leads the assert at the FontManager code.

**Example Case**

Screenshot_20230828-165244_TheXTech

```
00:45.591 [Fatal]: <Assertion condition has failed>:
---------------------------------------------------------
File: /home/runner/work/TheXTech/TheXTech/src/fontman/font_manager.cpp(536)
Function: textSize
Condition: g_fontManagerIsInit
---------------------------------------------------------
====Stack trace====
  # 0: 0x7943cbbe34  
  # 1: 0x7943cbc360  
  # 2: 0x7944074160  SDL_ReportAssertion
  # 3: 0x7943bd0f00  _ZN11FontManager8textSizeEPKcmijbj
  # 4: 0x794396ab3c  _Z15SuperTextPixLeniPKci
  # 5: 0x794386e250  _ZN11PauseScreen4InitEib
  # 6: 0x79438416d8  _Z9PauseGame9PauseCodei
  # 7: 0x79437b6e64  _Z8GameMainRK14CmdLineSetup_t
  # 8: 0x79439c84ac  SDL_main
  # 9: 0x794410f9d0  Java_org_libsdl_app_SDLActivity_nativeRunMain
```

**Additional context**

To resolve this issue, the new font manager should fall back to the legacy SMBX's font engine even calling FontManager's API directly. Otherwise, because of this bug, assets that lacks the `fonts` directory will don't work which is absolutely not intented to be: old assets should work as well as new according to goal of the TheXTech, updates of engine shouldn't require user to upgrade their assets to get the game work, but it should disable features that are depends on new-added resources.

Wohlstand:
This thing implements a legacy font engine as a part of modern font engine. So, when modern font fails to load for any reason (for example, when using old assets packages that lacks the `fonts` sub-directory), legacy fonts will take the place of new fonts. The new font manager will work as usually, but it will print using legacy built-in font textures.

Additionally, I made sure that translations won't load if fonts directory is missing. I tweaked the menu to gray out the languages item if fonts are missing.

Fixes #600

### DIFF
--- a/src/fontman/legacy_font.cpp
+++ b/src/fontman/legacy_font.cpp
@@ -1,0 +1,443 @@
+/*
+ * Moondust, a free game engine for platform game making
+ * Copyright (c) 2014-2023 Vitaly Novichkov <admin@wohlnet.ru>
+ *
+ * This software is licensed under a dual license system (MIT or GPL version 3 or later).
+ * This means you are free to choose with which of both licenses (MIT or GPL version 3 or later)
+ * you want to use this software.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ *
+ * You can see text of MIT license in the LICENSE.mit file you can see in Engine folder,
+ * or see https://mit-license.org/.
+ *
+ * You can see text of GPLv3 license in the LICENSE.gpl3 file you can see in Engine folder,
+ * or see <http://www.gnu.org/licenses/>.
+ */
+
+#include "sdl_proxy/sdl_stdinc.h"
+#include "legacy_font.h"
+#include "gfx.h"
+#include "std_picture.h"
+#include "core/render.h"
+#include "font_manager_private.h"
+
+
+LegacyFont::LegacyFont(int legacyFontId)
+{
+    loadFont(legacyFontId);
+}
+
+LegacyFont::~LegacyFont()
+{
+    m_charMap.clear();
+    SDL_memset(m_textures, 0, sizeof(m_textures));
+}
+
+void LegacyFont::loadFont(int fontId)
+{
+    m_charMap.clear();
+    SDL_memset(m_textures, 0, sizeof(m_textures));
+
+    switch(fontId)
+    {
+    default:
+    case 1: // Numbers font
+        m_glyphWidth = 18;
+        m_glyphHeight = 16;
+
+        for(int i = 0; i < 9; ++i)
+        {
+            char uchar[5] = {static_cast<char>('0' + i), 0, 0, 0, 0};
+            m_textures[i] = &GFX.Font1[i];
+
+            RasChar c;
+            c.texId = i;
+            c.tX = 0;
+            c.tY = 0;
+            c.tW = 16;
+            c.tH = 14;
+            c.width = 18;
+            c.valid = true;
+
+            m_charMap.insert({get_utf8_char(uchar), std::move(c)});
+        }
+
+        m_isReady = true;
+        m_fontName = "numeric";
+        break;
+
+    case 2: // World map level title font
+        m_glyphWidth = 16;
+        m_glyphHeight = 18;
+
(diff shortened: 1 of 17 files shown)


LoveBodhi:
Scr_2023-08-31_18-01-43

Without "Fonts" folder; OK, but still loads the graphics in the "i18n" folder

Scr_2023-08-31_18-02-34

With "Fonts" folder; OK

Tested with asset: Nostalgic Paradise

Wohlstand:
Ye, looks fine, thanks for the test!
Now I need @ds-sloth's code review until merge the thing.

LoveBodhi (approved):
But anyways: Test OK.

src/fontman/legacy_font.h
```diff
@@ -0,0 +1,135 @@
...
+    {
+        uint32_t texId = 0; //!< Texture index in m_textures array
+        uint32_t tX = 0; //!< Horizontal pixel offset on texture
+        uint32_t tY = 0; //!< Vertical pixel offset on texture
+        uint32_t tW = 0; //!< Pixels width of the texture fragment
+        uint32_t tH = 0; //!< Pixels height of the texture fragment
+        uint32_t width = 0; //!< Width of glyph
+        bool valid = false; //!< Is a valid glyph
+        int32_t x = 0;//!< X pixel offset
+        int32_t y = 0;//!< Y pixel offset
```
ds-sloth, line 115: Can be shrunk a lot (could be nice to do this at RasterFont as well).

```
struct RasChar
{
    uint8_t texId = 0;
    bool valid = false;
    uint16_t tX = 0;
    uint16_t tY = 0;
    uint8_t tW = 0;
    uint8_t tH = 0;
    uint8_t width = 0;
    int8_t x = 0;
    int8_t y = 0;
};
```

Size: 36 bytes -> 12 bytes

src/fontman/legacy_font.cpp
```diff
@@ -0,0 +1,443 @@
...
+        m_isReady = true;
+        m_fontName = "font2";
+        break;
+    }
+}
+
+PGE_Size LegacyFont::textSize(const char* text, size_t text_size,
+                              uint32_t max_line_lenght, bool cut,
+                              uint32_t /*fontSize*/)
+{
```
ds-sloth, line 201: How different is this from `RasterFont`'s implementation? I'd really encourage you to make `LegacyFont` inherit from `RasterFont` for code size and maintainability. (Same comment applies for later methods.)

src/fontman/legacy_font.h
Wohlstand: I guess, I also can get rid of x and y, they are actually unused.

src/fontman/legacy_font.cpp
Wohlstand: The difference is significant: the LegacyFont's is much simpler internally (in a case of the glyph alignment, size computation, etc.), and it's also different because it needs a totally different format of textures and other stuff.

ds-sloth: I see, then let's keep this part as-is. :+1:

src/fontman/legacy_font.h
Wohlstand: Done just now, please take a look.

Wohlstand:
> I see some opportunities for optimization of the RasChar structs

I made some, should be better now.

> In terms of logical correctness: have you tested the case where there is a fonts directory at an episode / level but no global fonts directory? That seems like the only possible bug here to me.

Do you mean the work in a case of custom fonts at level/episode providen? Lemme check that...

Wohlstand:
> do the legacy fonts get used again correctly after the level?

I verified just now: yes! Just because of my backup and restore system at my font manager is universal, it preserves the state of default fonts and their overridings (with no matter of their type), and once level/episode gets quit, custom fonts gets unloaded and initial mappings gets restored back.

ds-sloth:
Code looks good; should be ready to merge, I'll do so after the CI builds finish.

PR #619, opened 2023-10-05
Refactor for BlockHit()

ds-sloth:
This PR attempts to preserve the existing logic precisely, only refactoring the code to be easier to read.

This refactors the `BlockHit()` method for two purposes:
(1) Removes the dead NPC spawn code for ancient `Block_t::Special` values between 100 and 999. The cases 101, 104, and 201 could still exist in SMBX64, and code to recreate them is added to the main NPC spawn code section. PGE-FL should be modified (also after our 1.3.6.1 release) to no longer replace these values so that we can maintain compatibility with SMBX64.
(2) Cleaned up duplicated logic for replacing powers with a base power if a player was small.


Wohlstand:
> The cases 101, 104, and 201 could still exist in SMBX64

The SMBX64 already proceeds these ancient spawn codes between 101 and to 106 to spawn NPCs. By my memory, PGE-FL already replaces these ancient codes from files with modern codes.

Wohlstand:
EDIT: Yes, PGE-FL already [DOES](https://github.com/WohlSoft/PGE-File-Library-STL/blob/master/file_rw_lvl.cpp#L339-L386) this, so, the game never gets these ancient codes being written as-is. They already replaced with modern codes.

ds-sloth:
> PGE-FL already does this

That's a problem for compatibility with SMBX64. SMBX64 converts codes 102, 103, 105, and 106 to modern codes, but still uses the legacy logic for codes 101, 104, and 201 if they are present in the level file. For compatibility, PGE-FL should pass those three codes to the game (as -101, -104, and -201).

Otherwise, certain bugs will occur in SMBX64 and not in the main game. (I can try to construct some examples later if you like.)

src/blocks.cpp
```diff
@@ -535,9 +535,30 @@ void BlockHit(int A, bool HitDown, int whatPlayer)
...
+        else if(b.Special == 104)
+        {
+            C = NPCID_GRN_BOOT;
+            is_ancient = true;
+        }
+        else // b.Special == 101
+        {
+            C = NPCID_FODDER_S3;
+            is_ancient = true;
+        }
```
Wohlstand, line 561: This code, seems, incomplete, there are MANY ancient codes to be properly treated:

See this list as a hint: <https://github.com/WohlSoft/PGE-File-Library-STL/blob/master/file_rw_lvl.cpp#L354-L381>

Anyway, I'll try to make a demo to verify the work...

Wohlstand: Okay, I found the place where code gets converted, all fine.

Wohlstand:
Going to merge this now. Also, I want to make few clean-ups after all.

PR #635, opened 2023-11-03
Semi-solid slope fixes

SpencerEverly:
This draft pull request fixes the NPC semi-solid slope issue, making it similar to X2's fixes.

The reason why it's a draft is because I was thinking of rewriting how slopes are collided with players/NPCs, but it's uncertain on if I plan on doing this anytime soon. I'll need some help with it too, which yeah, that's how confusing it is for me.


Wohlstand:
Also, seems you need to properly bump submodules, they are conflicting right now :eyes:
At my mainstream all submodules are up to date, I made sure for that.

SpencerEverly:
> Also, seems you need to properly bump submodules, they are conflicting right now 👀 At my mainstream all submodules are up to date, I made sure for that.

Just fixed it after figuring that out. Hopefully that won't happen again...

ds-sloth:
I tested this change when you shared it in Discord, and I'm not sure that it is truly an improvement over the current behavior.

I agree with your judgement that, if we start this work, we should probably revise the entire slopes system. However, I'm not sure that this is an important change to make right now.

(As a side note, we can't even consider this until it's guarded by a compat flag.)

ds-sloth:
As mentioned in #647, we don't currently have the development resources to review outside logic contributions.

However, I think it *is* likely we will want to fix this eventually. I don't think that the solution you've proposed here is complete enough for us to use.

If you're interested in helping us learn more about semi-solid slopes and how to make them well, please continue experimenting with them. Make complete test scenarios and consider how these semi-solid slopes interact with things like:
- moving layers (including NPC-attached layers)
- walking NPCs
- projectiles at multiple angles, including thrown NPCs, fireballs, and projectile generator NPCs
- shallow slopes

Keep revising your implementation as you test this. I'd be very eager to see what you find.

ds-sloth:
Probably not in an efficient way, the slope physics are integrated in a lot of other systems so it would be best to fix them in the engine itself.

Issue #638, opened 2023-11-09
Multiple save files/slots (Over 3)?


LoveBodhi:

vitzie629:
If SMBX had four save states, it would be just like Super Mario All Stars and I would find that neat!

PR #639, opened 2023-11-09
Allow more save slots

Wohlstand:
This is a small update that allows game to have 5 save slots, and makes menu to be able to have any number of save slots: just need to change the constant to set a number of save slots and rebuild the game.

#638

### DIFF
--- a/src/global_constants.h
+++ b/src/global_constants.h
@@ -69,12 +69,6 @@ const int maxCreditsLines = 200;
 
 const int maxSaveSlots = 5;
 
-constexpr int c_menuSavesLength = maxSaveSlots + 2;
-constexpr int c_menuItemSavesEndList = maxSaveSlots - 1;
-constexpr int c_menuItemSavesCopy = maxSaveSlots;
-constexpr int c_menuItemSavesDelete = maxSaveSlots + 1;
-constexpr int c_menuSavesOffsetY = (maxSaveSlots - 3) * 30;
-
 const int maxWorldCredits = 100;
 
 const int maxYoshiGfx = 10;
--- a/src/global_constants.h
+++ b/src/global_constants.h
@@ -69,6 +69,12 @@ const int maxCreditsLines = 200;
 
 const int maxSaveSlots = 3;
 
+constexpr int c_menuSavesLength = maxSaveSlots + 2;
+constexpr int c_menuItemSavesEndList = maxSaveSlots - 1;
+constexpr int c_menuItemSavesCopy = maxSaveSlots;
+constexpr int c_menuItemSavesDelete = maxSaveSlots + 1;
+constexpr int c_menuSavesOffsetY = (maxSaveSlots - 3) * 30;
+
 const int maxWorldCredits = 100;
 
 const int maxYoshiGfx = 10;
--- a/src/global_constants.h
+++ b/src/global_constants.h
@@ -67,7 +67,11 @@ const int MaxSavedEvents = 200;
 
 const int maxCreditsLines = 200;
 
+#ifdef LOW_MEM
+const int maxSaveSlots = 3;
+#else
 const int maxSaveSlots = 5;
+#endif
 
 const int maxWorldCredits = 100;
 
--- a/src/global_constants.h
+++ b/src/global_constants.h
@@ -67,7 +67,7 @@ const int MaxSavedEvents = 200;
 
 const int maxCreditsLines = 200;
 
-const int maxSaveSlots = 3;
+const int maxSaveSlots = 5;
 
 constexpr int c_menuSavesLength = maxSaveSlots + 2;
 constexpr int c_menuItemSavesEndList = maxSaveSlots - 1;
--- a/src/main/menu_main.cpp
+++ b/src/main/menu_main.cpp
@@ -84,7 +84,7 @@ static SDL_atomic_t         loadingProgrssMax = {};
 static SDL_Thread*          loadingThread = nullptr;
 #endif
 
-static constexpr int c_menuSavesLength = maxSaveSlots + 2;
+// static constexpr int c_menuSavesLength = maxSaveSlots + 2;
 static constexpr int c_menuItemSavesEndList = maxSaveSlots - 1;
 static constexpr int c_menuItemSavesCopy = maxSaveSlots;
 static constexpr int c_menuItemSavesDelete = maxSaveSlots + 1;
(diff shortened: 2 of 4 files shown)


src/global_constants.h
```diff
@@ -67,7 +67,13 @@ const int MaxSavedEvents = 200;
...
 const int maxCreditsLines = 200;
 
-const int maxSaveSlots = 3;
+const int maxSaveSlots = 5;
+
+constexpr int c_menuSavesLength = maxSaveSlots + 2;
+constexpr int c_menuItemSavesEndList = maxSaveSlots - 1;
+constexpr int c_menuItemSavesCopy = maxSaveSlots;
+constexpr int c_menuItemSavesDelete = maxSaveSlots + 1;
+constexpr int c_menuSavesOffsetY = (maxSaveSlots - 3) * 30;
```
ds-sloth, line 76: Move these new constants into menu main header or cpp

src/main/menu_main.cpp
```diff
@@ -1344,7 +1344,7 @@ bool mainMenuUpdate()
         else if(MenuMode == MENU_SELECT_SLOT_1P || MenuMode == MENU_SELECT_SLOT_2P)
         {
             if(SharedCursor.Move)
-                s_handleMouseMove(4, MenuX, MenuY, 300, 30);
+                s_handleMouseMove(c_menuItemSavesDelete, MenuX, MenuY - c_menuSavesOffsetY, 300, 30);
```
ds-sloth, line 1347: Multires sets MenuY to 30 on very low res screens. Instead of adding menu offset Y to MenuY, use it in GetMenuPos when calculating MenuY for the save select menu modes.

src/main/menu_main.cpp
```diff
@@ -1414,7 +1414,7 @@ bool mainMenuUpdate()
                     else
                     {
                         MenuMode -= MENU_SELECT_SLOT_COPY_S1_ADD;
-                        MenuCursor = 3;
+                        MenuCursor = maxSaveSlots;
```
ds-sloth, line 1417: logically, c_menuItemSavesCopy

src/main/menu_main.cpp
```diff
@@ -1812,7 +1812,7 @@ static void s_drawGameSaves(int MenuX, int MenuY)
     const auto& info = SaveSlotInfo[MenuCursor + 1];
 
     int infobox_x = ScreenW / 2 - 240;
-    int infobox_y = MenuY + 145;
+    int infobox_y = MenuY + 145 + c_menuSavesOffsetY;
```
ds-sloth, line 1816: This usage of c_menuSavesOffset here suggests that you're centering the menu on its old coordinates, not shifting it up. Make sure that the info box stays in the same location when entering the 1P char select screen.

src/main/menu_main.cpp
```diff
@@ -2077,18 +2077,18 @@ void mainMenuDraw()
...
     else if(MenuMode == MENU_SELECT_SLOT_1P || MenuMode == MENU_SELECT_SLOT_2P) // Save Select
     {
-        s_drawGameTypeTitle(MenuX, MenuY - 70);
-        SuperPrint(SelectWorld[selWorld].WorldName, 3, MenuX, MenuY - 40, 0.6f, 1.f, 1.f);
-        s_drawGameSaves(MenuX, MenuY);
-        XRender::renderTexture(MenuX - 20, MenuY + (MenuCursor * 30), GFX.MCursor[0]);
+        s_drawGameTypeTitle(MenuX, MenuY - (70 + c_menuSavesOffsetY));
+        SuperPrint(SelectWorld[selWorld].WorldName, 3, MenuX, MenuY - (40 + c_menuSavesOffsetY), 0.6f, 1.f, 1.f);
+        s_drawGameSaves(MenuX, MenuY - c_menuSavesOffsetY);
+        XRender::renderTexture(MenuX - 20, MenuY + (MenuCursor * 30) - c_menuSavesOffsetY, GFX.MCursor[0]);
```
ds-sloth, line 2083: Another part where it'll be much clearer if you move the logic into GetMenuPos

ds-sloth (changes requested):

src/global_constants.h
Wohlstand: Good note! I'll do!

ds-sloth:
Let's see, the three main issues are (1) RAM use, (2) filesystem use, and (3) UI design.

(1) Should be under a kilobyte but I can check this to confirm.
(2) Maybe more concerning, remember that on devices with slow I/O the menu hangs while gamesaves are checked. I can try to look more closely at this soon, there might be a way to optimize this.
I don't remember what the relative cost of a dirlist vs a file existence check is on 3DS / Wii but if a dirlist is much cheaper, we could use a DirListCI here.
(3) You should adjust MenuY itself instead of adding an offset in all places. The right procedure to modify is GetMenuPos.
It might be better to scroll this menu than to expand it. The current design results in the menu header moving, which is unfortunate since I was reworking the char select screen to keep the menu header. The current design also may not work at low resolutions.

src/main/menu_main.cpp
Wohlstand: I added this because I done the shift of the whole menu block, and I keeping info box being sticked to the bottom of the shifted menu.

ds-sloth:
Another note: Krissy suggested splitting the menu into pages of 3 (not sure if vertical or horizontal scrolling). This seems like a very good idea to me and it'd totally fix the UI issues noted here.

Wohlstand:
1) Every save slot (at the static array with structs and stuff), shouldn't be so big, the rest of all data gets loaded dynamically and use the same amount of data with no dependency on the number of save slots. The save slots array supposed to keep only data are being used to display and use in the menu.

2) It's always possible to set different constant value for different targets, that's no problems at all. I.e. keep the constant to be equal 3 on too low-end platforms, and others can have 5.

3) The case of lower resolutions feels like need to give a slightly different design of menu to keep it compact.

src/global_constants.h
Wohlstand: Done.

src/main/menu_main.cpp
```diff
@@ -2083,42 +2089,48 @@ void mainMenuDraw()
...
-        SuperPrint(SelectWorld[selWorld].WorldName, 3, MenuX, MenuY - (40 + c_menuSavesOffsetY), 0.6f, 1.f, 1.f);
-        s_drawGameSaves(MenuX, MenuY - c_menuSavesOffsetY);
+        MenuY -= c_menuSavesOffsetY;
+
+        s_drawGameTypeTitle(MenuX, MenuY - 70);
+        SuperPrint(SelectWorld[selWorld].WorldName, 3, MenuX, MenuY - 40, 0.6f, 1.f, 1.f);
+        s_drawGameSaves(MenuX, MenuY);
 
-        SuperPrint(g_mainMenu.gameEraseSlot, 3, MenuX, MenuY - 30 + (5 * 30), 1.0f, 0.7f, 0.7f);
+        SuperPrint(g_mainMenu.gameEraseSlot, 3, MenuX, MenuY - 30 + (c_menuSavesLength * 30), 1.0f, 0.7f, 0.7f);
```
ds-sloth, line 2131: MenuY + (c menu delete item * 30)

Wohlstand: No, all fine, the "c_menuItemSavesDelete" is `maxSaveSlots + 1`, and "c_menuSavesLength" is `maxSaveSlots + 2`, there are different values.

Wohlstand: EDIT: Oh, wait! the `-30` been removed, that makes sense!

ds-sloth: Ah, I see, this is a different pattern; feel free to ignore

ds-sloth:
Everything looks fine. If you want to try a scrolling implementation I think that would keep the menu design more consistent, but fine if not. Only remaining note is that the MenuY adjustments should occur directly in GetMenuPos so that they can be easily tweaked at low resolutions.

Wohlstand:
> directly in GetMenuPos so that they can be easily tweaked at low resolutions.

I'll try to tweak that...

Wohlstand:
Done!

Wohlstand:
Speaking about the unified metadata file (for menu only) is better to implement after Medal Tracking merge. I am in process of reviewing the code.

ds-sloth (approved):
Everything looks fine currently, we should confirm that the UI looks good in multires and with the char select screen before the next release.

PR #643, opened 2023-11-14
Implemented support for Wii U consoles

Wohlstand:
This is mostly SDL-based thing. The [SDL2 for WiiU](https://github.com/devkitPro/SDL/tree/wiiu-sdl2-2.28) is very completed and very fresh.

Anyway, the thing still needs more testing before this can be considered as a completed thing.


Wohlstand:
There are next known bugs was found during tests:

On CEMU (by me):
- Happens on Release builds (MinSizeRel): When completing a level, it's impossible to move playable character, possibly because of a controller connection lost. Re-plugging (setting "OFF" and then "ON" in settupgs of CEMU) temporarily escapes the problem.

On Hardware (tested by `@Bubble` on Matrix):
- Can't write anything to SD card (and therefore, unable to save the game, setting, and unable to write log files at all).

LoveBodhi:
> Happens on Release builds (MinSizeRel): When completing a level and letting new paths open, it's impossible to move playable character on the world map, possibly because of a controller connection lost. Re-plugging (setting "OFF" and then "ON" in settupgs of CEMU) temporarily escapes the problem.

This means I cannot save the game and quit after level clear.

Wohlstand:
**About bugs:**
- Buggy work of release builds in the CEMU actually is a bug of CEMU itself: it's recompiler is buggy, and there is a way to run the game correctly: running with the `--force-interpreter` command-line argument.
- The inability to write SD card had been fixed: actually was a bug of `mkpath` implementation at the DirManager that confused everything.
- Since recent time, the same build of the game should work with SD card as well as USB sticks, or as a part of WUHB packages.

Wohlstand:
**Also:**
- As `@Bubble` tested on her hardware, the scaling algorithm change doesn't crash the game. I should try out the same via interpreter thing...

Wohlstand:
Last to do stuff left:
- [ ] Add custom worlds / battles directories from SD cards and USB sticks when game runs inside the WUHB package.

ds-sloth:
I've read everything and all looks fine to merge on my side. I don't fully understand the path logic in `app_path_wiiu.cpp` but since you'll be the port maintainer for Wii U that's not a problem. It's unfortunate that you need the `#ifdef` in the default controls section, but understandable.

If we want to improve the port's experience a bit, it could be nice to add full Wiimote support using the same Wiimote support code from Wii. Unfortunately it seems that the function names on Wii and Wii U are different, not sure if the semantics are different also.

Wohlstand:
That `#ifdef` is a kind of a workaround, just, because on SDL2 side, the location of A/B/X/Y supposed to have the same physical position as on XBox controller. Even if it's a Switch Pro controller.

ds-sloth (approved):

PR #647, opened 2023-11-23
Add missing global memory values for LunaDLL and the future XTechScript system

SpencerEverly:
In this draft pull request, I decided to add a couple more memory addresses from 1.3 over to this pull request.

The reason why I am doing this is because whenever the Lua scripting system starts development, we'll already have all the necessary global memories to not only use for the future scripting system, but also for my future project to emulate LunaLua (LunaXTech, more details coming when Lua script development gets started).

This pull request will no longer be a draft once most memory addresses are put together.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,7 +1,3 @@
-[submodule "3rdparty/PGE_File_Formats"]
-	path = 3rdparty/PGE_File_Formats
-	url = https://github.com/WohlSoft/PGE-File-Library-STL.git
-	branch = master
 [submodule "3rdparty/LuaJIT"]
 	path = 3rdparty/LuaJIT
 	url = https://github.com/WohlSoft/LuaJIT.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -44,6 +44,3 @@
 	path = 3rdparty/FileMapper
 	url = https://github.com/WohlSoft/FileMapper.git
 	branch = master
-[submodule "3rdparty/PGE_File_Formats"]
-	path = 3rdparty/PGE_File_Formats
-	url = https://github.com/WohlSoft/PGE-File-Library-STL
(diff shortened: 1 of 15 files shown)


Wohlstand:
Don't forget to revert the `.gitmodules` into initial state, this change has no sense. Also, missing the branch name.

ds-sloth:
I've discussed our roadmap with the other main developers and we've decided that right now the project doesn't need game logic changes from outside contributors. We're focusing a lot on finalizing the base engine upgrades and the scripting system.

We find your work interesting and encourage you to keep working on your TheXTech fork, including this X2 compatibility layer. Please ask about it in code help, show it off in showcase, and if your contributions line up with our goals, we'll work with you to mainline them after finishing version 1.3.7. But we don't currently have the development time to review all of this work or integrate it with the main project's goals and constraints.

Concretely, I took a look at this PR. I understand and respect the goals here but I want to share some very quick thoughts:
- A lot of the new source code makes reference to outside material. (For instance, using "rupee" to describe the coin type known as "gem" in the engine's source code.) We only include such references when they come from Redigit's code or comments and are in the process of removing them to make the engine asset-neutral.
- You extend mememu a lot but it's likely that mememu will not be used *at all* in the TheXTech Lua API. If you would like your X2 compatibility layer to create something like mememu, you should plan to do so in pure Lua. (But, this decision isn't finalized yet.) The mememu data structure is actually very costly and I'd like to avoid adding fields that are not used by existing SMBX 1.3 Autocode content.
- The contributed source code makes certain things configurable just because they happened to have settable memory in SMBX 1.3 (such as the different NPC IDs for gems spawned in different contexts). Special cases like the ones you've added will not be a part of the TheXTech Lua API, but we plan to make the API flexible enough that it would be possible to rewrite the code for spawning specific NPCs from grass, pods ("eggs" in SMBX asset language), or blocks.

We'll update the PR guidelines to help clarify these considerations and others soon.

Issue #402, opened 2022-08-29
Fancy Character Selection for Multiplayer


0lhi:
One subject discussed in the Abstract Controls Pull Request that was kept for later is the idea of a fancier Multiplayer Screen:

![Multiplayer](https://user-images.githubusercontent.com/74732674/187184565-2c8c95a0-a077-46c1-87de-1e65bff0f177.png)

> @0lhi The Power-Up status of the Character should be indicated. The selected Characters should play the running animation.

> @ds-sloth (1) The screen is ugly: yes. I picked it this way to make it as similar to the experience of the original game as possible but allowing the two players to select their characters at the same time. Your design looks fantastic. I can try to make something like it. Do you think it should be used by default? Always or just for multiplayer? There are also some uncertainties in your proposal. What should happen when player 1 is selecting a character and player 2 is selecting a profile?

> @0lhi
>> (1) Do you think it should be used by default?
>
>Yes!
>
>> Always or just for multiplayer?
>
>Just for Multiplayer. Singleplayer game should presume that the the controls are already setup to the Players liking, and stick with the Classic Character Selection. This is actually important for Mode 2 & 3, since Menu Quits are allowed in Speedruns.
>
>> There are also some uncertainties in your proposal. What should happen when player 1 is selecting a character and player 2 is selecting a profile?
>
>Move the Profile and Testing Controls Menu a bit down, and lock Player's 2 Cursor inside it, while Player 1 can still navigate freely.

ChristianSilvermoon:
I actually would prefer to see the new character selection for single player too, both for the sake of consistency **and** seeing the power up state ahead of time.

It's not uncommon that I'll use different characters to 'store' power-ups in a run, so seeing it ahead of time could save me switching later. I'd strongly prefer it.

0lhi:
Counterpoints:

1. A strong visual distinctions will make Players immediately clear when they're in the wrong Screen (a problem that I've been having)
2. It would be a too big deviation from other SMBX Versions
3. It might invalidate Mode 2+3 Runs as it'd give Players an information advantage.

The Single Player Selection shouldn't be touched.

I wouldn't mind making a 1P Game accessible from the Multiplayer Screen, however.

ChristianSilvermoon:
> A strong visual distinction will make Players immediately clear when they're in the wrong Screen (a problem that I've been having)

This sounds more like a separate readability issue :P

> It would be a too big deviation from other SMBX Versions

[...]

PR #664, opened 2023-12-11
Implement save slot pages

ds-sloth:
Here I've implemented a page-based menu for save slot selection following Wohlstand's work for #638.

This keeps the menu a uniform height and the menu header in the same position, which is aesthetically pleasing.

The non-primary pages are currently difficult to discover but this is an advanced feature, so I think that is okay. We also have some left / right scroll indicator icons coming when I implement #402. I can add those indicators to this menu at that time.

I like the UI and think this is ready to merge from a user-facing perspective (but would be happy to hear feedback). The main thing possibly blocking merge is that the numerical logic is difficult (because every page has the Copy / Erase options). @Wohlstand, does it look okay to you, or should I try to beautify it some.

Warning: as always, be cautious when testing a branch affecting the save game logic. People have lost save games in the past. I don't expect anything to be broken here, but you should not test this build if you would be upset to lose the copy of your save games in its folder.


ds-sloth:
These changes have been deferred until after the Config Update is merged. The save slot limit has been temporarily reverted to 3.

ds-sloth:
I'm moving this from v1.3.7 to v1.3.7.x, and might implement it together with some changes to the episode select design. (Not totally certain yet.)

Wohlstand:
Just now I paid attention on the GIF demo, and first in my mind that pages were supposed to be explicitly indicated, otherwise it's non-intuitive.

ds-sloth:
This is outdated for a few reasons, including the unintuitive issue, and the left right Playstyle switching. In the next few weeks I'll implement a different tab based approach based on designs by Krissy.

PR #716, opened 2024-02-06
Added sub-hubs mechanism

### LINKED ISSUE #641: Implement the sub-hubs system
Sub-hubs - levels are considered as part of main hub, but they were done as separated levels.

There are two known examples I can give:
- "Sarasaland Adventure 2" by Otabo (packed game).
- "The Boo Adventure 2" (Episode for SMBX).

Currently, player must quit the sub-hub to get ability to save the level. Also, if player takes an exit at the third-layer level (a level entered inside sub-hub), player will quit from the main hub at unexpected entrance (see #196, vanilla bug).

## To resolve this, the sub-hubs concept is needed:
- Add a mark to each level "is sub-hub" that will be handled by the engine.
- When player appears at sub-hub, it should be remembered separately from the main hub.
- When player loses their character at sub-hub, player will resume at last entered door ot at start point of sub-hub.
- When player loses their character at level entered from the recent sub-hub, player should enter the recent sub-hub from the last entered door (the same door as being used to enter the level).
- Player must be able to save the game while at sub-hub.
- When resuming game from the main menu, player must appear at the sub-hub instead of the main hub.

Wohlstand:
This thing allows to make sub-hubs in episodes. To use this, the [update of TheXTech SDK](https://github.com/Wohlstand/TheXTech-SDK/commit/13e08e029f6f871592cd04f8ef7137e0fd8d4ded) is needed.

- ~~Every level will have a JSON field at level-wide extra settings.~~
- ~~Level that has the "is_subhub" being set to `true` will allow player to save the game here, and load it back at here, not from the main hub.~~
- World maps will have a JSON world-wide extra settings that lists all levels that will work as sub-hubs.
- Level that are listed in the world map's sub-hubs list will allow player to save the game here, and load it back at here, not from the main hub.
- Sub-hubs will work even on World-Map based episodes: player can save at the sub-hub, and resume game at sub-hub with no matter it's a world map based episode.

Imperfection on the moment right now:
- If going into the new hub from another hub or sub-hub, it's required to go via warps. Attempt to go into sub-hub without the target warp will lead player enter from a same warp id as being remembered at the previous hub.

Fixed #641


src/game_main.cpp
```diff
@@ -1057,6 +1062,9 @@ int GameMain(const CmdLineSetup_t &setup)
             if(LevelRestartRequested && Checkpoint.empty())
                 StartWarp = lastWarpEntered; // When restarting a level (after death), don't restore an entered warp on checkpoints
 
+            if(IsHubLevel && StartWarp > 0) // Save the warp where player entered the hub
+                ReturnWarpSaved = StartWarp;
+
```
ds-sloth, line 1067:
```
else if(IsHubLevel && ReturnWarp == 0)
    ReturnWarpSaved = 0;
```

src/globals.cpp
```diff
@@ -302,6 +303,7 @@ std::string FileNamePathWorld;
 int FileFormatWorld = 0;
 
 bool IsEpisodeIntro = false;
+bool IsHubLevel = false;
```
ds-sloth, line 306: Please remove the old `IsHub()` definition; I think it's later in this file

src/main/level_file.cpp
```diff
@@ -199,6 +199,25 @@ bool OpenLevelData(LevelData &lvl, const std::string FilePath)
...
+    if(IsEpisodeIntro)
+    {
+        IsHubLevel = NoMap;
+        FileRecentSubHubLevel.clear();
+    }
+
+    // Level-wide extra settings
+    if(!lvl.custom_params.empty())
+    {
+        const nlohmann::json level_data = nlohmann::json::parse(lvl.custom_params);
```
ds-sloth, line 209: Please check for exceptions here (the `try` block should also include the `value` check, that throws if there is a type mismatch)

src/main/screen_pause.cpp
```diff
@@ -157,7 +157,7 @@ static bool s_QuitTesting()
 
 static bool s_SaveAndContinue()
 {
-    bool CanSave = (LevelSelect || (IsEpisodeIntro && NoMap)) && !Cheater;
+    bool CanSave = (LevelSelect || (IsEpisodeIntro && NoMap) || IsHubLevel) && !Cheater;
```
ds-sloth, line 160: Replace with simply `(LevelSelect || IsHubLevel)` (and, scan the whole repo for all other cases of `IsEpisodeIntro && NoMap`)

ds-sloth (changes requested):
I shared specific comments in the code. I can make some of the fixes myself if that's easier for you.
General notes:
* This appears to allow hub levels on world map episodes. That's fine, even good, but is this intentional?
* What situations (death? level complete? etc) should return to the main hub (or world map), and what situations should return to the sub-hub?
* As a related note, you don't seem to remember `ReturnWarp` for the main hub after entering a level from a sub-hub. Is the idea that the main hub should only be returned to by a warp out of the sub-hub?

src/game_main.cpp
Wohlstand: Good note!

ds-sloth: Actually, a further note: at this point in the code, should `ReturnWarp` refer to the sub-hub, or the main hub?

src/globals.cpp
Wohlstand: Done, I'll send soon

src/main/level_file.cpp
Wohlstand: Ouch, forgot, good catch!

src/main/screen_pause.cpp
Wohlstand: Done, I replaced all such cases as `IsHubLevel` replaces that condition.

ds-sloth:
From our discussion about possible cases, it's becoming clear that this is a more complex feature than I had thought.

I agree with @0lhi's recommendation that it is necessary for us to move all of the information about sub-hubs into the WLDX file. I know that the UI for inputting this is more complex, but it should be worth it if it will make future decision-making easier.

I've shared an example episode here based on the flowchart we were discussing. The doors in the center column represent hubs / sub-hubs, and the doors on the right column represent levels. [Sub-Hub World Map.zip](https://github.com/Wohlstand/TheXTech/files/14186251/Sub-Hub.World.Map.zip)

For me to approve this PR, I'll need:
(1) the three of us to review and approve the behavior of the Hub World and World Map test cases I shared (one of us can update them when we move the sub-hub tracking into the WLDX) 
(2) you to move the sub-hub tracking from the LVLX into the WLDX. It should be fine to add a `std::vector<std::string> WorldSubHubLevels;` of filenames to TheXTech's globals.
(3) confirmation that we haven't altered compatibility in any way for an episode with no sub-hubs present

0lhi:
I'd like the implementation keep these future possibilities open (without making a statement on whether they should become a thing):

1. Allowing saving and loading at checkpoints
2. Allowing to use any warp as a checkpoint/savepoint
3. Allowing to save in every level for certain episodes

I also think the feature should be renamed from "Subhub" to "SaveLevel", as there can be different reasons for wanting to save mid-level (example, very long Kaizo-ish stage).

ds-sloth:
I don't think that the feature should be renamed, because a user may be confused if taking a warp from a "Save Level" to another level and then dying returns them to the Save Level.

As an additional note, I don't think remembering checkpoints or internal warps inside of a sub-hub is in the scope of this PR (but the possibility is open).

ds-sloth:
Updates on my merge criteria:
(1) We still need to review and approve the behavior of the Hub World and World Map test cases I shared (**updated here**: [Sub-Hub World Map.zip](https://github.com/Wohlstand/TheXTech/files/14299003/Sub-Hub.World.Map.zip))
(2) Done by the two of us
(3) Let's review the code changes, and make sure nothing is different if `SubHubLevels` is empty and `enable_last_warp_hub_resume` is off

For 0lhi's suggestions, I don't know whether they are in scope for this PR. If we want to do any of these as part of this PR, we need more test cases.

* Allowing to save in every level for certain episodes shouldn't be hard to add if desired. We shouldn't reuse `IsHubLevel`, so that we can allow exiting a saved level that is not a sub hub normally.
* Separately, should we save Checkpoint index as part of `ReturnWarpSaved`? We could easily do this by using negative indexes, if desired. We should do this **only** for the "recent sub-hub".
* Thirdly, should we save sub-hub internal warps as part of `ReturnWarpSaved`? Again, we should do this **only** for the "recent sub-hub".

0lhi:
> For 0lhi's suggestions, I don't know whether they are in scope for this PR. If we want to do any of these as part of this PR, we need more test cases.

Want to bring my attention to `(without making a statement on whether they should become a thing)` :P

I don't think it's in scope.

ds-sloth (changes requested):
I just read all of the code for compatibility, and it looks safe to me. Assuming that we will not add any more features to this PR, then the only thing left is testing.

I have tested the example I shared, and I think everything works as intended in the Hub World example. In the World Map example, it seems that dying or getting an exit always returns the player to the World Map.

I thought we had planned for the player to return to the most recent sub hub, at least if they die or get an exit in a level (and possibly also if they die or get an exit in a hub).

Wohlstand:
I going to test my studf on my side too, and other things. Oh  one another thing that I forgot: there are possible warps that may be hidden initially (for example, the final boss door that gets open with a noisy animation). And I am not sure how player will enter back the level. Possibly, such warps are hidden initially should being remembered and never register them as "entrance", and instead, use the previous available warp or player start point instead.

Wohlstand:
An example of such boss door is the "2k15 Summer Takeover" which appears at the pre-last last sub hub. Also, there are subhubs were possible to be enabled at the "Mario + Luigi' Marathon" I guess, and... more levels set as sub-hubs or even enable the "save everywhere" 🤔

ds-sloth:
> there are possible warps that may be hidden initially

Good note, but let's file that as a separate issue, it affects all episodes when the compat flag `enable_last_warp_hub_resume` is set.

ds-sloth:
As a note, there is some precedent for using offscreen exit to return to the world map from a sub-hub, "2k15 Summer Takeover" does this. I think that we should return to the world map after offscreen exits and deaths (if "retry on fail" is off) in sub hubs.

Wohlstand:
> Good note, but let's file that as a separate issue, it affects all episodes when the compat flag enable_last_warp_hub_resume is set.

About this warp, it's visible, so, a wrong example
изображение

Wohlstand:
This warp is also visible :thinking: 
изображение

ds-sloth:
To be clear, my requested change is:
* Make sure that at world map worlds, dying or collecting an exit in a level returns to the sub-hub (if present). This is not the case in the current build.
* While doing this, make sure that it remains the case that at world map worlds, dying or collecting an exit in a sub-hub itself returns to the world map.

Wohlstand:
Merging now.

PR #717, opened 2024-02-06
In-game asset pack selection!

ds-sloth:
## After a lot of discussion and design work, we have an implementation of in-game asset pack selection!

Scr_2024-02-06_17-14-00

### Setup:
* Extract your favorite asset packs to to `assets/` inside your installation directory.
* For non-portable installations, you may also extract them to `~/.PGE_Project/thextech/assets/` on Linux or `C:\Users\<username>\.PGE_Project\thextech\assets\` on Windows.
* **IMPORTANT**: to be recognized, your asset pack must include its ID in `gameinfo.ini`. We will add `id="smbx"` to the `[game]` section of SMBX's `gameinfo.ini`, and `id="aod"` to Adventure of Demo's `gameinfo.ini`.
* Advanced: if you have two versions of the same asset pack, you should add `version="..."` to at least one of them. Otherwise, only one of them will be detected.
* Asset packs stored directly to your `thextech` folder will still be loaded, but this location is deprecated and kept only to make migration easier. The game will internally treat them as having a blank asset pack ID.
* You do **NOT** need to install an additional copy of your current asset pack unless you want to migrate your installation.

### Use:
* Hold the Select (Drop Item) button at the title screen to choose an asset pack. You may need to be patient, since the game will fully reload using the new asset pack.
* Advanced: launch the engine with `-c <asset-pack-id>` or `-c <asset-pack-id>/<asset-pack-version>` to specify which asset pack you want to run from the command line, independent of install location.
* Note for advanced users: launching the engine with `-c <path-to-asset-pack>` now requires the target asset pack to specify an ID, and you will need to migrate your user content to subdirectories with that ID.

### Data management:
* Now, you will see subdirectories for each asset pack in your `worlds/` and `settings/gamesaves/` folders (as well as the other folders).
* You can place your episodes and find your other data here.
* All data is shared between different versions of the same asset pack.
* Why do I have two copies of my favorite asset pack? If you installed a pack to both your main `thextech` directory and to `assets/`, then it will be recognized as two separate asset packs. After you migrate all of your data to `worlds/<id>/`, `battle/<id>/` and `settings/gamesaves/<id>/`, you may remove the original asset pack.

## Review and testing

I believe that this feature is fully complete and polished, but there is always room for improvement. @Wohlstand, please review the code and make sure that it seems appropriate to you.

For usability testers, please start by confirming that the instructions above work well for your existing installation and everything works as you expect. Then, if you have the time, please try other installation types (portable, non-portable with assets installed in the same directory as the executable, non-portable with assets stored in your user directory). Thank you!


ChristianSilvermoon:
Okay so to sum up my experience and all questions/concerns...

## Migration
Migration from the former 'mono-assets' system to this new asset pack system isn't overly complex, but does have a variety of steps

For the most part it appears that it's a matter of
1. Create `assets/PACK` (whatever is appropriate for your legacy asset pack)
2. Move `graphics`, `sound`, `sounds.ini`, `music`, `music.ini`, `introset` (if available), `intro.lvl`, `outro.lvl`, and very importantly `gameinfo.ini` to this new location
3. Move the episodes and battle levels that came with the episode by default to 'assets/PACK/worlds' and `assets/PACK/battle` respectively as well
4. Move non-default episodes and battle levels to `worlds/PACK` and `battle/PACK` respectively
5. Move all saves located under `settings/gamesaves` to `settings/gamesaves/PACK`
6. Add `id=PACK` under the `[main]` key of `assets/PACK/gameinfo.ini`

While this process isn't complicated, it is a six step process that some users may incorrectly follow, leading to confusion when the game doesn't find their save data anymore.

I would recommend (although from past discussions this idea seems like it may not be desirable) providing an option to migrate automatically, setting the `version` key to "legacy" or "migrated"

If no built-in migration system is included, then either a seperate utiltiy or script, or (possibly the simplest option) the recommendation to download a fresh copy of the game and instructions on what to copy where from the old install

## Concerns about MenuGFX2.png
A number of episodes in the past have shipped custom sets of assets intended to overwrite the original game's core assets; for multiple reasons we're trying to avoid that particular setup now.

Many of these include a custom `intro.lvl` or `MenuGFX2.png` that won't be seen in the event things are simply copied to the episode directory...

I am also unsure if `outro.lvl` placed in the episode directory would be detected and used by the game.

Some do not include any custom `MenuGFX2.png` at all, leading to either a missing logo, or a logo that looks identical to another asset pack.

### Suggestion
I think it's important that the `id` be visible *somewhere*, at least during selection. I know it may not be the most visually appealing, but in the event that two different packs have the same `MenuGFX2.png` it can become difficult to tell them apart, leading to user frustration. It can be somewhere unobtrusive ideally. 

I am concerned that such asset packs are a very real possibility. It should probably be suggested that it is bad practice to re-use the logo unless it's a pack variant with a unique version number, but for users simply converting older episodes with assets to asset packs, I cannot honestly see that *not* being an issue for someone at some point with no other indicator.

For episode directories containing a `MenuGFX2.png` I think it would be nice if the logo changed when the episode was selected, but I could also understand if this poses an issue on lower-spec platforms

I suggest displaying the `id` string because it should always be unique and should not be localization.

## Conclusion
The new system feels perfect to me, with the exception of the concern I have ^^

This is a very nice and substantial upgrade over the old way of doing things and makes things more flexible for Linux/UNIX packaging formats, and easier for users to switch out packs without having to restart the game or maintain multiple copies of the game.

ds-sloth:
Hi, thanks a ton for this review. We (the core team) discussed your concern about MenuGFX2 seriously, and want to suggest that:
- In general, a packed game that does not change the title logo should probably be converted into an episode instead of an asset pack
- If a converter wants a packed game to become an asset pack, they should update `MenuGFX2.png` manually, or specify a different logo image at `[android] logo = "graphics/..."`.
- But, we could possibly add a `show-id` flag to the `[game]` section, which would force the pack's ID to be shown on the selection screen. Do you think that would be necessary?

Separately, I'll note: we got a suggestion on Matrix to animate the left/right transition between asset packs in the selection screen. I'll try to do this soon.

ds-sloth:
I added a `[main] show-id = 1` flag for `gameinfo.ini`. Can you please confirm that it works as expected for your content?

As a side note, I also revised the animation when a logo is overridden to make the logo fade in/out (instead of having it suddenly change as it moves up and down).

ChristianSilvermoon:
This works, I suppose, but I honestly don't understand why this shouldn't be a global setting, and editing `gameinfo.ini` per pack to add that when they're in .ISO format or other formats like that is going to by a lot more difficult and time consuming.

May I ask what the opposition to having it be a global setting is?

ds-sloth:
I think we must have a slightly different understanding of what the role of asset packs is. In my opinion, asset packs are not like episodes; there's an expectation that the person creating them knows what they are doing. If a person creates an ISO asset pack with a non-unique logo, and doesn't set `show-id`, I think that is a defective asset pack and you should complain to the person who created it.

We are already planning to add code to Moondust Maintainer to allow converting packed games into episodes; we can definitely add the ability to set the ID, and the ability to enable `show-id`, when we are doing so. Since ISO-creation software is somewhat rare, adding these checks to Maintainer should be quite effective.

I definitely wouldn't want this to be a global setting, simply because we're trying to limit those as much as possible to avoid too much clutter in the Config Update.

ChristianSilvermoon:
> I think we must have a slightly different understanding of what the role of asset packs is. In my opinion, asset packs are not like episodes; there's an expectation that the person creating them knows what they are doing. If a person creates an ISO asset pack with a non-unique logo, and doesn't set `show-id`, I think that is a defective asset pack and you should complain to the person who created it.
> 
> We are already planning to add code to Moondust Maintainer to allow converting packed games into episodes; we can definitely add the ability to set the ID, and the ability to enable `show-id`, when we are doing so. Since ISO-creation software is somewhat rare, adding these checks to Maintainer should be quite effective.
> 
> I definitely wouldn't want this to be a global setting, simply because we're trying to limit those as much as possible to avoid too much clutter in the Config Update.

I think we the same understanding as to their role, just different concerns.
I think the solution of having `show-id` in `gameinfo.ini` is probably good enough in this case, you're right.

ds-sloth:
I'm glad! Anything else you'd like me to revise before we call this feature "done"? (I think Wohlstand and 0lhi still need to approve it as well.)

ChristianSilvermoon:
> I'm glad! Anything else you'd like me to revise before we call this feature "done"? (I think Wohlstand and 0lhi still need to approve it as well.)

I think it's perfect, although, I've noticed for some reason the Lowser's Conquest logo does not seem to animate... wonder what could cause that? :thinking:
[Scr_2024-02-10_05-29-57.webm](https://github.com/Wohlstand/TheXTech/assets/30162808/bee09774-8f37-44ce-a56c-472976b34694)

Oh, I guess GitHub won't embed webms? ~~Shame on me for trying to be nice about file sizes I guess XD~~

ds-sloth:
That's correct; since Lowser's conquest's logo is not actually MenuGFX2, it would fail to animate correctly at resolutions lower than 800x600. You can check out a7ee0bb to see if you like, this was fixed in 597f0d2.

0lhi (approved):

ds-sloth:
Yes, you're right, it seems that 1280x720 was the only resolution that animated perfectly. Here's what 800x600 looked like, and lower resolutions are even worse (due to `intro.lvl` being exceptionally poorly optimized for multires).

**OLD SCREENSHOTS:**
 


Yes, the menu graphics are provided by `intro.lvl` and NOT present in `MenuGFX2.png`.

In general, if the logo is overridden, I don't think it's safe to assume that the override logo is present in the menu screen at the same location as MenuGFX2 would normally be, so I'm happy with using the fade animation instead of the "slide" animation.

ChristianSilvermoon:
> Yes, you're right, it seems that 1280x720 was the only resolution that animated perfectly. Here's what 800x600 looked like, and lower resolutions are even worse (due to `intro.lvl` being exceptionally poorly optimized for multires).
> 
> **OLD SCREENSHOTS:**  [  ](https://private-user-images.githubusercontent.com/72112344/303961189-945f5835-8e7f-44dd-a023-0aa463dd944d.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MDc3MDI5NzUsIm5iZiI6MTcwNzcwMjY3NSwicGF0aCI6Ii83MjExMjM0NC8zMDM5NjExODktOTQ1ZjU4MzUtOGU3Zi00NGRkLWEwMjMtMGFhNDYzZGQ5NDRkLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDAyMTIlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwMjEyVDAxNTExNVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTRkYjkzYjI3YWVhNWRmNmRhNzdlODcyNTQ2Mjk3NzgzOTE3Y2ZhNzFmMzFkM2Y5YzE1MTQyZmIyZDRiMmRjY2ImWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.vP-I950rQMa__T01KfTyGXmfTIH5z6DkU17pbWuLsAE) [ ](https://private-user-images.githubusercontent.com/72112344/303961189-945f5835-8e7f-44dd-a023-0aa463dd944d.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MDc3MDI5NzUsIm5iZiI6MTcwNzcwMjY3NSwicGF0aCI6Ii83MjExMjM0NC8zMDM5NjExODktOTQ1ZjU4MzUtOGU3Zi00NGRkLWEwMjMtMGFhNDYzZGQ5NDRkLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDAyMTIlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwMjEyVDAxNTExNVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTRkYjkzYjI3YWVhNWRmNmRhNzdlODcyNTQ2Mjk3NzgzOTE3Y2ZhNzFmMzFkM2Y5YzE1MTQyZmIyZDRiMmRjY2ImWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.vP-I950rQMa__T01KfTyGXmfTIH5z6DkU17pbWuLsAE)  [  ](https://private-user-images.githubusercontent.com/72112344/303961225-eaf53a01-ed9e-4be1-8470-0e196b8e6e37.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MDc3MDI5NzUsIm5iZiI6MTcwNzcwMjY3NSwicGF0aCI6Ii83MjExMjM0NC8zMDM5NjEyMjUtZWFmNTNhMDEtZWQ5ZS00YmUxLTg0NzAtMGUxOTZiOGU2ZTM3LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDAyMTIlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwMjEyVDAxNTExNVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWFiNjZiZWRlN2MwNTJkZTgyZGE4MzE4OGY0MDNlNmE0YWM2NmM4YTZmMTI2NTcxNzlhZWMxODgwM2YzNzQ1ZTYmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.wNuF_12sQL9wGepao7u1zgLtbAFo04i61Lc99ecfNWE) [ ](https://private-user-images.githubusercontent.com/72112344/303961225-eaf53a01-ed9e-4be1-8470-0e196b8e6e37.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MDc3MDI5NzUsIm5iZiI6MTcwNzcwMjY3NSwicGF0aCI6Ii83MjExMjM0NC8zMDM5NjEyMjUtZWFmNTNhMDEtZWQ5ZS00YmUxLTg0NzAtMGUxOTZiOGU2ZTM3LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDAyMTIlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwMjEyVDAxNTExNVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWFiNjZiZWRlN2MwNTJkZTgyZGE4MzE4OGY0MDNlNmE0YWM2NmM4YTZmMTI2NTcxNzlhZWMxODgwM2YzNzQ1ZTYmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.wNuF_12sQL9wGepao7u1zgLtbAFo04i61Lc99ecfNWE)
> 
> Yes, the menu graphics are provided by `intro.lvl` and NOT present in `MenuGFX2.png`.
> 
> In general, if the logo is overridden, I don't think it's safe to assume that the override logo is present in the menu screen at the same location as MenuGFX2 would normally be, so I'm happy with using the fade animation instead of the "slide" animation.

Ah, I see, that makes sense

src/main/asset_pack.cpp
```diff
@@ -0,0 +1,480 @@
...
+        ret.logo_override = true;
+
+    // don't allow any dirsep characters in asset pack ID
+    for(char& c : ret.id)
+    {
+        if(c == '/')
+            c = '_';
+
+        if(c == '\\')
+            c = '_';
```
Wohlstand, line 133: This can be simplified with `else` branch:
```cpp
if(c == '/' || c == '\\')
    c = '_';
```

src/main/asset_pack.cpp
```diff
@@ -0,0 +1,480 @@
...
+
+            // strip the ID from a legacy asset pack
+            if(!is_custom_root)
+            {
+                if(!pack.id.empty() && !pack.version.empty())
+                    pack.version = pack.id + '-' + pack.version;
+                else if(!pack.id.empty())
+                    pack.version = pack.id;
+
+                pack.id = "";
```
Wohlstand, line 186: Why not to use the `.clear()` here?

src/main/asset_pack.cpp
```diff
@@ -0,0 +1,480 @@
...
+}
+
+static inline void s_load_image(StdPicture& dest, std::string nameWithPng)
+{
+    dest.reset();
+
+#if defined(X_IMG_EXT)
+    if(nameWithPng.size() > 4)
+    {
+        size_t base_size = nameWithPng.size() - 4;
```
Wohlstand, line 63: Would be safer if use the:
```cpp
size_t base_size = nameWithPng.find_last_of('.');
if(base_size == std::string::npus)
   base_size = nameWithPng.size();
```

src/main/asset_pack.h
```diff
@@ -0,0 +1,82 @@
...
+
+    std::unique_ptr<Gfx> gfx;
+
+    bool show_id = false;
+    bool show_version = false;
+    bool logo_override = false;
+
+    bool operator==(const AssetPack_t& o) const
+    {
+        return this == &o;
```
Wohlstand, line 59: Are you sure this works? Or I miss something... I would add a comparison with every field by field, and return false when any is not equal.

Wohlstand (approved):
Okay, I finished the review, and seems everything is fine for me. I did a small fix by myself. Other things I'll probably tweak soon. I need to update some on Haiku and other platforms. We should start the move away from `~/.PGE_Project/thextech` into `~/.thextech` (with keeping use old if old directory exists do don't ruin everything).

src/main/asset_pack.h
ds-sloth: This was being used in a single place to delete an asset pack from the list by identity. I moved the logic to the calling site instead for clarity.

PR #719, opened 2024-02-13
Added an option: Don't grab shells from top fall

Wohlstand:
This option disables an ability of player to grab a shell while stomping it from the top and holds the RUN button. It's very importand option to prevent soft-locks in ported legacy episodes where player can accidentally destroy a shell hidden in the niche which is required to continue the path.

## Default behaviour:
```ini
[compatibility]
allow-grab-from-top-fall = true
```
Scr_2024-02-13_23-25-41

## When option is disabled:
```ini
[compatibility]
allow-grab-from-top-fall = false
```
Scr_2024-02-13_23-25-31

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -28,6 +28,7 @@ Changelog for 1.3.7-dev
 Changelog for 1.3.6.5-dev
 -Fix TheXTech 1.3.6 bug where auto movement on the world map was not reset when switching episodes (@ds-sloth)
 -Fix TheXTech 1.3.6 bug where dropping a player at the world map removed their mount (@ds-sloth)
+-Added a compat.ini option to disable the grabbing of the shell from the top falling to resolve soft-locking problems in early SMBX episodes (@Wohlstand)
 
 Changelog for 1.3.6.4
 -Use OpenGL as the default renderer on supported platforms ("sdl" may be used in config.ini or at at the command line to request the SDL2 renderer)
--- a/src/compat.cpp
+++ b/src/compat.cpp
@@ -126,6 +126,8 @@ static void compatInit(Compatibility_t &c)
     c.fix_frame_perfect_despawn = true;
     // 1.3.6.3
     c.pound_by_alt_run = true;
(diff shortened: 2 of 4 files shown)


ds-sloth (approved):
I read everything, it looks clean, and should have a minimal impact on performance, and it seems useful for making old episodes playable.

It solves a number of cases of destroying essential NPCs, which get solved by Generators in TI2.

Let's wait for @0lhi to look quickly at this before merging.

0lhi:
I am very hesitant about this one. You know I'm short on time these days so for now I'll just ask:

How do you plan to prevent soft-locks by spin-jumping shells?

Wohlstand:
> How do you plan to prevent soft-locks by spin-jumping shells?

That's already possible by Autocode, it's need to make the `TurnAround` player field being always `false`. It's the way how the `Superb Demo Sisters` disables the Spin-jump ability.

0lhi:
I don't think this flag belongs in the Engine. While we agreed to include select Pre-1.3 features on a case-by-case basis, this flag adds a way too situational change in game-logic that has little benefit and will catch users off-guard.

A proper fix in this situation would be to do what The Invasion 2 does in the Brick Hole Massacre Level - use a generated Shell that hides/spawns it's own generator.



For greater immersion a new Generator Type "Pouf" could be added. It would provide a cool feature for new content, fix this issue, and even render the Spin Jump ban obsolete.

Wohlstand:
The purpose of this change to DON'T modify the TI1 in the set of "SMBX Nostalgie" game pack, and this change was done exclusively for TI1 and probably this thing won't be documented.

There is not just one level, there is MORE of them:

Brick niche:
Scr_2024-02-14_00-43-23

Bonus level:
Scr_2024-02-14_00-44-32

The case that you mentioned, it is not related to this thing:
Scr_2024-02-14_00-45-20

Wohlstand:
Okay, as this idea is BAD, I'll close this at all. This feature can be added, but NOT in this form as it was done here.

PR #731, opened 2024-02-23
Fixed the visual buck on star-locked pipe enter

Wohlstand:
This is a small vanilla bug when a wrong frame gets shown when player attempts to enter the star-locked pipe by down direction.

I made sure the thing will work correctly with any character with any state.

Demo level to test behaviour: [Starred pipes test.zip](https://github.com/TheXTech/TheXTech/files/14381996/Starred.pipes.test.zip)

## Fixed behaviour:
Scr_2024-02-23_08-00-46

## Original behaviour:
Scr_2024-02-23_08-03-45


ds-sloth:
Just tested locally, it seems to work well. The code itself is very elegant, nice work. I will quickly guard behind a new multipurpose compat flag, `fix-visual-bugs`, for two reasons: (1) it's _possible_ that the player frame during the pause state somehow affects the game logic, and (2) it's important for users to get exactly the original experience when in Mode 3. Then I'll approve.

ds-sloth (approved):
Looks fine to me. Very clean fix. You will see that I added a compat flag for this. Feel free to merge at your convenience.

PR #732, opened 2024-02-23
New user paths for TheXTech 1.3.7

ds-sloth:
The following notes reflect my and @Wohlstand's decisions for the 1.3.7 paths update. Please see the edit history for more information about the history of this PR.

- [x] In all cases except legacy compat, `TheXTech` refers to the game prefix set in CMake.
- [x] `THEXTECH_FIXED_ASSETS_PATH` and `THEXTECH_USER_DIR_NAME` are deprecated and will not be used.

# Linux and other UNIX-like systems (except macOS)

- [x] New path: `$XDG_DATA_HOME/TheXTech/` 
- [x] Keep the compatibility with old `~/.PGE_Project/thextech/` directory when it exists to don't break the stuff.
- [x] Special DEB packages that had custom `/.thextech-smbx/` and `./thextech-aod/` directories should be auto-migrated by BASH scripts that will work as overlays over engine, and they should leave symbolic link for uninformed users and for dumb users who ignore any importand messages and then asking dumb questions.
- For the case of system-installable games:
  - [x] Have the separated package of engine that will depend on architecture, and game package that will be "no-arch" and will depend on engine's package.
  - [x] Put every individual asset pack at the `/usr/share/games/TheXTech/<pack name>/`. Game will automatically search for assets at these directories and allow successful launch with `thextech -c <pack-id>`.
  - [x] Every game package should include the Desktop file, and the bash script as an overlay that should migrate legacy user directories (like `~/.thextech-smbx/` and `~/.thextech-aod/`) and maintain other backend stuff.
  - [ ] Multi-action Desktop file will allow users to override the system-wide asset pack (by installing to `$XDG_DATA_HOME/TheXTech/assets/`), with a "safe mode" option to force load the system-installed asset pack.

# macOS case

- [x] Keep the `~/TheXTech Games/` a hive of user directories for bundled games only (the name of each game gets generated from the template: `~/TheXTech Games/<name of bundle without .app ending>`).
- [x] For the case of plain engine and a pile of externally unpacked game packs, use the `~/TheXTech/` directory. Every individual game pack should be placed in the `~/TheXTech/assets/<pack name>/` directory.
- [x] Old `~/TheXTech Games/Debug assets/` will be abandoned in favour of the new directory

# Windows case

- [x] It's similar to Linux case, for the compatibility, use the `%UserProfile%\.PGE_Project\thextech\` as a fallback.
- [x] As a new location for plain engine use, use `%APPDATA%\TheXTech` (resolves to `%UserProfile%\AppData\Roaming` on Vista+ and `%UserProfile%\Application Data` on Windows XP) for the new stuff.

# Haiku case

- [ ] When making packages, have two variants of packages: game engine itself, and the game pack that depends on engine's package and has all the shortcuts.
- [x] When installing gamepacks, put them at the `/boot/system/data/TheXTech/<pack name>/`
- [ ] User directory should be `$HOME/config/settings/TheXTech/` and when `$HOME` is undefined (empty), fall back to the constant `/boot/home/`.

# Android case

- The case of manually selected game pack directories should work normally as usually.
- **SPLIT THE FOLLOWING TO ANOTHER ISSUE**:
  - The case of fallback directory (currently at the `/Android/data/ru.wohlsoft.thextech/...`) should be revised, as it makes problems for Android 11+ users who can't access that directory to backup/restore gamesaves. Additionally, it was made dirty so it produces collisions when running namesake episode from different game packs.
  - At the application directory (used as fallback when assets directory is read-only), make the sub-directory `<app data dir>/gamesaves/<pack-id>` to save new gamesaves instead of the united hive and resolve the collision between game packs problem. Old gamesaves should be still readable for compatibility reasons.
  - Implement an ability to backup and restore gamesaves from the application directory as users are unable to access that directory by file managers since Android 11+.
  - Do use the `<Phone>/TheXTech/` directory instead of gamepacks's self directories for all the userdata like logs, gamesaves, etc.
  - When `<Phone>/TheXTech/` directory exists, search for presented gamepacks in `<Phone>/TheXTech/assets/` and automatically add them into list. These packs will always presented and never been removed on clean action. They should be shown in the game selection menu together with manually-added game packs.

# Emscripten case

- Currently all assets are packed into an archive by Emscripten build, and userdata is separated between packed games. This should be fine to keep if we want.
- In the long future, we will be able to download asset packs from a repo within the engine, and then we can use a desktop-like system.
- Split this to another issue: but we also need ways for users to upload / download their own content (asset packs, episodes, gamesaves). Use `libarchive` and upload/download tools here.

# Consoles

- [x] Rename `thextech` to `TheXTech` on homebrew platforms other than Vita and Switch.
- [ ] The rename can be done automatically, but make the notification about this action.
- [x] Wii U: packed games now use the main user paths, treating the packed assets as a built-in asset pack. Gamesaves and settings are now stored on external storage if possible. (Confirm with Wohlstand that this is desirable.)


Wohlstand:
Small note for the Haiku:
- The root directory to save user settings (and gamesaves) would go into the `$HOME/config/settings/TheXTech` (fall back to `/boot/home` if `$HOME` is undefined).
- And the directory for user-placed game packs and custom worlds would go into `$HOME/TheXTech`

Other cases I trying to figure by [reading this article](https://www.haiku-os.org/docs/develop/packages/BuildingPackages.html) (where is would be a "system-wide installed" assets hive, i.e. when every assets pack is a package) :thinking:

Wohlstand:
I guess, that could be:
```
/boot/system/data/TheXTech/
```

ds-sloth:
I'd advise keeping things as simple as possible on Haiku. Would keeping the whole user directory at `$HOME/TheXTech` be acceptable? `/boot/system/data/TheXTech` sounds pretty good to me, but I'm not a haiku user so my instincts are not developed.

ChristianSilvermoon:
On the topic of AppImages, it's worth noting that while they are neat and useful they're sometimes caught up in weird situations, such as the transition from libfuse2 to libefuse3

They are rarely simple to update, among some other things, but I do think they'd be good to support for the sake of convenience and completeness, and allowing users to choose what kind of packaging they'd like.

In short, I'd love to see them as an option, but would be wary of DEB packages being phased out in favor of them.

ds-sloth:
I don't think we are planning to phase out DEBs at all (and if we were, the closest competitor would be flatpak, not AppImage), but I am interested in possibly phasing out our statically-linked Ubuntu TAR builds in favor of AppImages.

ChristianSilvermoon:
> I don't think we are planning to phase out DEBs at all (and if we were, the closest competitor would be flatpak, not AppImage), but I am interested in possibly phasing out our statically-linked Ubuntu TAR builds in favor of AppImages.

Ah, perfect then, I'd say that's a good idea personally.

In that case if a user were not, or could not directly use the AppImage, then the option of
```bash
TheXTech.AppImage --appimage-extract
```
would still be available to them to extract it

Wohlstand:
About the Windows, I have something to append: soon I can make a kind of installers using the InnoSetup thing, so, TheXTech also can appear as installable thing (can appear at `%ProgramFiles%` as bin installation, and the assets directory). I'll clarify this moment later, this message is just an idea (for now I should go out of home).

Wohlstand:
About packages, I managed to build the CI that produces DEB packages from every maintained assets pack and providing the template for everything (to produce a similar DEB package just using a custom thing).

ds-sloth:
I'm going to merge this now for some brief testing in the main branch prior to the 1.3.7-beta release.

PR #734, opened 2024-02-26
Config Update

ds-sloth:
### Features:
* A new options menu accessible from the title screen...<br>
* and in-game, including easy user configuration for all of the onscreen information options offered by TheXTech<br>
* The options menu excludes confusing compatibility options that could break episodes (these are locked behind a code, `edityourfriends`)<br>
* An easy option at the title screen to disable TheXTech bugfixes and updates that may impact compatibility<br>
* and the ability to start speedruns from the game menu, without any command-line options (by hitting select at this prompt)<br>
* A new Speedrun Mode 2 that is defined as including the bugfixes essential for NetPlay and multiplayer support, in addition to the bugfixes included in traditional Mode 2. (Note that this means that speedrunners interested in "authentic" SMBX 1.3 behavior should be using Mode 3, which has always been the case.)<br>

### Internal Features:
* A unified compatibility-config structure that uses a simple, declarative notation to define options.
* An internal notion of option types that allows options irrelevant to gameplay to be freely changed in compatibility, speedrun, and NetPlay modes.
* Full translation support for in-game options (but, no translations have been created yet, and these should wait until we have finalized the set of options)

### Todo before merge:
* Decide how WLDX files will indicate that they target TheXTech 1.3.7+ (and therefore should default to Modern Mode on), and implement this at Moondust and this branch. [DONE: WLDX will default to Modern Mode, and WLD will default to Classic Mode]
* Review all settings and make sure that non-essential settings get removed, that all settings are named and placed appropriately, and that only the necessary settings can be altered by creators at `compat.ini`.
* Review the legacy settings migration code **after** all other tasks are done, and ensure it is still accurate.


SkullGamer205:
Interesting feature.

LoveBodhi:
I'm assign for Config Translation purpose.

Wohlstand:
> Decide how WLDX files will indicate that they target TheXTech 1.3.7+ (and therefore should default to Modern Mode on), and implement this at Moondust and this branch.

This could being added as an extra setting that can be set via Moondust Editor or something :thinking:

0lhi:
My notes so far:

1. On existing saves, it would make sense to have "Select" skip the second page and enable Vanilla directly.
2. moneytree can get you back to 0, which should not be the case.
3. Imo it's fine to always recommend Modern for WLDX. The chance of breaking stuff is minor, and even so, 1.3.7 is allowed to make such changes for the sake of longevity.

ds-sloth:
Thanks a lot for the notes!

1. That should help reduce user confusion, since the Speedrun modes aren't allowed to be selected on an existing save. I'll do that as soon as possible.
2. This doesn't really have to do with the Config Update, but it's a good note. If we don't want `moneytree` to "redeem" players who have thousands of negative coins, it could be nice to make it only go up to the next hundred coins (as it currently does for positive coins).
3. Seems like a great idea to me. A bit simpler than using a new WLDX JSON setting. @Wohlstand, what do you think about this?

0lhi:
Life/Debt Gameinfo doesn't adjust to Mode of Savegame:

Normal:

Scr_2024-03-02_20-19-14

Menu Mode 3:

Scr_2024-03-02_20-19-33

Also quitting a Mode 3 Speedrun enables Menu Mode 3, which locks out of enabling Non-Mode 3 Games. Menu Mode 3 should probably be command-line exclusive 🤔

0lhi (approved):
Very happy with how it turned out ^-^

Wohlstand:
Going to deeply review it soon, right now I feel sleepy and slightly ill.

LoveBodhi:
Segmentation fault
When I pressing down in the controls menu, Segmentation fault has been occurred!!!
[TheXTech_log_2024_03_09_07_56_55.txt](https://github.com/TheXTech/TheXTech/files/14544189/TheXTech_log_2024_03_09_07_56_55.txt)

src/config.h
Wohlstand, line 1: Sounds like, this header needs to be completely eliminated as it has nothing than one single inclusion of another header. :thinking:

src/config.h
```diff
@@ -19,99 +19,927 @@
...
+            CreatorLocal = CreatorEpisode | CreatorFile,
+            Creator = CreatorGlobal | CreatorLocal,
+
+            Episode = UserEpisode | CreatorEpisode,
+            File = UserFile | CreatorFile,
+            Local = Episode | File,
+            Global = UserGlobal | CreatorGlobal,
+            All = 0xff,
+        };
+    };
```
Wohlstand, line 67: Maybe use the `enum class` here? C++11 introduces this feature here.

ds-sloth: `enum class` doesn't support bitwise operations. But, I think I can simply this whole system now, with just three values (user-settable, assets-settable, episode-settable).

src/config.h
```diff
@@ -19,99 +19,927 @@
...
+        SCALE_DOWN_SAFE = 1,
+        SCALE_DOWN_ALL = 2,
+    };
+
+    opt_enum<int> scale_down_textures{this,
+        {
+            {SCALE_DOWN_NONE, "none", "None", "Least loading stutter"},
+            {SCALE_DOWN_SAFE, "safe", "Safe", "Checks if images are in 2x format"},
+            {SCALE_DOWN_ALL, "all", "All", "Less loading stutter than 'Safe'"},
+        },
```
Wohlstand, line 548: What if mention the RAM usage too?

LoveBodhi:
Should an option, called "Sound Output Device", be configurable? (e.g. WASAPI, DirectSound, etc.)

Wohlstand:
Typically this is not needed until serious case like default API is glitchy on certain machine.

ds-sloth:
Do you mean that the "mono" setting doesn't work on Windows? If so, @Wohlstand, is there a way to force MixerX to mix everything to mono? If there isn't, let's simply remove the option on Windows since it doesn't work.

Wohlstand:
In a condition if WASAPI output is used, when DirectSound or some, then it should work. WASAPI doesn't allow to select custom spec, and therefore the thing always used obtained settings.

src/config/config_legacy_settings.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+    ini->beginGroup("sound");
+    if(ini->hasKey("disable-sound"))
+    {
+        bool audio_disabled;
+        ini->read("disable-sound", audio_disabled, false);
+        audio_enable = !audio_disabled;
+        audio_enable.m_set = level;
+    }
+    if(ini->hasKey("sample-rate"))
```
Wohlstand, line 70: Please add spaces between closing brace and the next independent `if` condition. It's not `else`.

src/config/config_legacy_settings.cpp
```diff
@@ -0,0 +1,139 @@
...
+#ifndef NO_WINDOW_FOCUS_TRACKING
+    if(ini->hasKey("background-controller-input"))
+    {
+        bool val;
+        ini->read("background-controller-input", val, true);
+        background_work = val;
+        background_work.m_set = level;
+    }
+#endif
+    ini->endGroup();
```
Wohlstand, line 116: Looks like it's a duplicated `endGroup` that refers nothing :thinking:

src/config/config_legacy_settings.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+        int val = -1;
+
+        ini->readEnum("compatibility-mode", val, -1, compatMode);
+        if(val != -1)
+        {
+            compatibility_mode = val;
+            compatibility_mode.m_set = level;
+        }
+    }
```
Wohlstand, line 136: Didn't you miss the `ini->endGroup();` here at the bottom? :eyes:

src/config/config_legacy_settings.cpp
Wohlstand: Okay, I fixed this by myself.

src/main/screen_options.cpp
```diff
@@ -0,0 +1,1022 @@
...
+inline void s_set_dirty()
+{
+    global_dirty = true;
+}
+
+inline void s_switch_section()
+{
+    if(section_index == SECTION_NONE || section_index >= g_options.m_options.size())
+        return;
+}
```
Wohlstand, line 87: Looks like this function makes nothing than checking the condition and it doesn't return anything :eyes:

What the true sense of this function right now?

src/config.h
```diff
@@ -556,6 +542,23 @@ class _Config_t
...
+    opt_enum<int> audio_channels{this,
+        {
+            {1, "mono", "Mono", nullptr},
+            {2, "stereo", "Stereo", nullptr},
+        },
+        defaults(g_audioDefaults.channels), {}, Scope::User,
+        "audio-channels", "Channels", nullptr,
+        config_audio_set};
+#else // #ifndef _WIN32
+    static constexpr int audio_channels = 2;
```
Wohlstand, line 559: Just a note: while default requests the 2, it's possible when the 5.1 or 7.1 output has been configured, then it will use 6 or 8 channels actually (such number of channels will be reported via Obtained, and will be used de-facto).

src/config.h
```diff
@@ -19,99 +19,930 @@
...
+#endif // #ifndef _WIN32
```
Wohlstand, line 560: Okay, right now the thing is broken: I forgot about some:
- This filed must be writeable, not constexpr (otherwise build fails as there are some places do store defaults on load).
- It's still necessary to allow edit this field, BUT, restrict by manual edit of INI file only, not by GUI.

src/core/sdl/sdl_core.cpp
```diff
@@ -26,10 +26,13 @@
...
 {
+    (void)(setup);
+
     bool res;
 
-    if(setup.allowBgInput)
-        SDL_SetHint(SDL_HINT_JOYSTICK_ALLOW_BACKGROUND_EVENTS, "1");
+    // if(g_config.background_work)
+    // apply this unconditionally -- otherwise, need to restart game for background-work to function as expected
+    SDL_SetHint(SDL_HINT_JOYSTICK_ALLOW_BACKGROUND_EVENTS, "1");
```
Wohlstand, line 35: This thing is an option, because I bet, there are some people who don't want it being backgroud input at all. For example, some other software can also use gamepad too and then cause the game to be poked unintentionally. Anyway, this thing needs some more info to make a proper choice on this.

src/core/sdl/window_sdl.cpp
```diff
@@ -198,6 +197,13 @@ bool WindowSDL::initSDL(uint32_t windowInitFlags)
     s_emscriptenFillBrowser();
 #endif
 
+#if RENDER_FULLSCREEN_ALWAYS // Use a full-screen on Android & PS Vita mode by default
```
Wohlstand, line 200: Looks like this comment needs to be extended: there are more platforms than just Android and Vita which use the full-screen only mode: 3DS, Wii, Wii U, Switch, etc.

src/effect.cpp
```diff
@@ -594,6 +600,7 @@ void UpdateEffects()
                     if(e.Frame >= 4)
                         e.Frame = 0;
                 }
```
Wohlstand, line 602: Whoops, this block needs to be slightly moved left, otherwise, a mess in the code style...

src/frm_main.cpp
```diff
@@ -314,10 +318,8 @@ bool FrmMain::restartRenderer()
     bool try_gl = false;
 
     CmdLineSetup_t setup;
-    setup.renderType = g_videoSettings.renderMode;
-    setup.vSync = g_videoSettings.vSync;
 
-    if(setup.renderType == RENDER_ACCELERATED_AUTO || setup.renderType == RENDER_ACCELERATED_OPENGL || setup.renderType == RENDER_ACCELERATED_OPENGL_ES || setup.renderType == RENDER_ACCELERATED_OPENGL_LEGACY || setup.renderType == RENDER_ACCELERATED_OPENGL_ES_LEGACY)
+    if(g_config.render_mode == Config_t::RENDER_ACCELERATED_AUTO || g_config.render_mode == Config_t::RENDER_ACCELERATED_OPENGL || g_config.render_mode == Config_t::RENDER_ACCELERATED_OPENGL_ES || g_config.render_mode == Config_t::RENDER_ACCELERATED_OPENGL_LEGACY || g_config.render_mode == Config_t::RENDER_ACCELERATED_OPENGL_ES_LEGACY)
```
Wohlstand, line 322: Maybe break this into 4 lines instead of a very-very long line?

src/main.cpp
```diff
@@ -369,7 +376,7 @@ int main(int argc, char**argv)
                                                    "  smbx2  - Disables all features and bugfixes except fixed at SMBX2\n"
                                                    "  smbx13 - Enforces the full compatibility with the SMBX 1.3 behaviour\n"
                                                    "\n"
-                                                   "  Note: If speed-run mode is set, the compatibility level will be overridden by the speed-run mode",
+                                                   "  Deprecated: acts as an alias for speed-run mode. Will be overridden if speed-run mode is set.",
                                                     false, "modern",
                                                    "modern, smbx2, smbx3",
                                                    cmd);
```
Wohlstand, line 382: The main reason that I added the separated compat level parameter is enabling the compatibility mode without forcing user to enable the speed-runner specific mode that adds the mode indication and speedrun timer. That's the main difference of the compat-level and speed-run mode.

Wohlstand: EDIT: Ouch, I forgot that you implemented the ability of user to choose the compat mode in the menu per episode :eyes:

Wohlstand (commented):
I finished the reviewing of the thing and seems to be fine. I going to test this out on my Mac Mini G4, and debug the Wii U problem. On PPC Mac right now I had some troubles to build it because of missing OpenGL functions and I'll try some soon.

Wohlstand:
Okay, going to merge this. Now.

PR #838, opened 2024-09-16
Cheat shadowstar now uses 75% black instead of 100%

ds-sloth:
Why do this?
* Original behavior was invisible against a black background.

Why not allow the old behavior in compat mode?
* This is a cheat code and normally used for debugging purposes only.
* The render code is extremely hot and it would slow the game down on low-end targets to check a compat flag here.


0lhi (approved):
Looks great! ^-^

Scr_2024-09-17_11-51-30

Wohlstand:
I checked this, looks fine for me!
Also, maybe backport this to the stable branch too?

ds-sloth:
I don't think so, because it's a minor change, and it will be more difficult to implement with the previous color format (4 x float).

Wohlstand (commented):
Forgot to say: everything looks fine for me!

PR #860, opened 2024-11-18
Implemented the in-game assert failure message

Wohlstand:
On platforms that doesn't have SDL's message box implemented, show the in-game assertion error message until close the game. Debugged primarily on Wii U, need to test on Vita, 3DS, Switch, and Wii.

This is required to avoid undescribed dead freezes and crashes without of any explanation left, just because of no message box implementation on these platforms.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1311,7 +1311,12 @@ elseif(PGE_MIN_PORT)
     target_compile_definitions(thextech PRIVATE -DLOW_MEM)
 endif()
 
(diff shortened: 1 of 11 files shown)


ds-sloth:
Tested on my devices, these changes seem very good to me.

Some minor changes on my side:

* I moved the exit hook into the message screen's logic (instead of the pause screen's logic). That is important because of some work I'm doing (currently in a private branch) to fix the game's nested mainloop issue.
* I added some more macro checks to reduce codesize on platforms without this feature.
* I removed your debug counter.

This is ready to merge on my side, you may want to check on the remaining homebrew platforms before merge:

- [x] Wii U
- [x] 3DS
- [ ] Wii
- [ ] Vita
- [ ] Switch

ds-sloth (approved):

PR #903, opened 2025-01-30
Add sound thread to do non-blocking SFX

### LINKED ISSUE #830: AoD: Bullet bills lower fps
### Describe the issue

In the level Dungeon of Agony the bullet bills that pop into existance will lower the fps to were I can die.

### A type of bug

Native bug

### Does this issue happens always or randomly?

Always happens

### Version

1.3.7-dev

### Version Hash

#590838a

### Platform

3DS

### Processor architecture

ARM32 / ARMv7 (32-bit ARM)

### Log file (if presented)

_No response_

### Example Case

_No response_

### Recording

_No response_

### Vanilla Recording

_No response_

### Additional context

_No response_

ds-sloth:
This puts the SFX play calls into a separate thread. The audio might be slightly choppy in these cases, but the gameplay will no longer be slowed down when the audio is delayed.

Uses `SDL_cond`, so we should check that the game still works on all targets before merging this. **Testing should include changing the MixerX settings in Advanced Settings**

- [x] Linux desktop
- [ ] macOS desktop
- [ ] Windows desktop
- [ ] Android
- [x] Emscripten
- [x] 3DS
- [x] Wii
- [ ] Wii U
- [ ] Switch
- [ ] Vita
- [ ] Haiku

Should resolve #830

### DIFF
--- a/src/sound_thread.cpp
+++ b/src/sound_thread.cpp
@@ -1,0 +1,166 @@
+/*
+ * TheXTech - A platform game engine ported from old source code for VB6
+ *
+ * Copyright (c) 2009-2011 Andrew Spinks, original VB6 code
+ * Copyright (c) 2020-2025 Vitaly Novichkov <admin@wohlnet.ru>
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef THEXTECH_NO_SDL_BUILD
+
+#include <vector>
+#include <cstdint>
+#include <exception>
+
+#include <SDL2/SDL_thread.h>
+#include <SDL2/SDL_mutex.h>
+
+#include "sound.h"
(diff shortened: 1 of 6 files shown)


Wohlstand:
As I already suggested in the Discord yesterday, it's better to set the threaded controlling on next conditions:
- On any platforms, the sound buffer set is longer than 1024 (for example)
- The target has is a low-end (3DS and Wii, possibly Vita too).

ds-sloth:
I'm not sure what negative effects this could possibly have. It's not good design for the main thread to need to wait for the mixer lock to be released, on any platform.

Wohlstand:
My main worry is that this might increase latency of sound effects (on some Android devices and some other platforms, latency is already visible especially when jumping, etc.)

ds-sloth:
(Just rebased on the new main branch -- this was necessary so that we can successfully cherry-pick this PR to the stable branch after merge)

src/sound_thread.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+#else // #ifndef THEXTECH_NO_SDL_BUILD
+
+// fallback: just call directly
+
+#include "sound.h"
+
+void PlaySfx(int Alias, int loops, int volume, uint8_t left, uint8_t right)
+{
+    PlaySfx_Blocking(Alias, loops, volume, left, right);
```
Wohlstand, line 159: It's possible to simplify this: just have a macro when no-SDL build is active and disable building this code file.

Wohlstand (approved):
I had some thinking about this, including the whole logic of the stuff, and I got an idea of the better solution: rather than running a kind of separate thread, make a kind of "queue" inside the MixerX itself:
- It will use the same thread as audio.
- Queue will have its own mutex that should be independent from the normal audio locker.
- Audio thread will lock the queue's mutex for a very short time to pick up the entry.

Until that, this thing feels like a workaround, but seems it's fine for now. I tested this on Linux, but not yet on other targets.

PR #909, opened 2025-02-13
Added option for real full-screen mode

### LINKED ISSUE #842: Use true full-screen mode on legacy hardware
Currently, game uses the Desktop fullscreen mode which doesn't supposes to switch the real resolution. On the rest of devices this works fine. However, legacy platforms that has weak GPU and CPU too (as an example, Mac Mini G4 that has Radeon 9200 video card, and game render lags when running big resolution, like 1920x1080 or even just 1280x1024, and legacy Windows computers like Pentium III computer that may have any of legacy video chips that won't carry larger resolutions quicker). And therefore, these platforms do need the ability to turn into the true fullscreen with the resolution switching. It's possible to have a startup detection of characteristics or have the specs hardcoded for platforms that doesn't need a detection, and then, use the true fullscreen or the desktop fullscreen for each of these cases.

Wohlstand:
This mode supposed to toggle real screen resolution, primarily needed to increase performance on slower devices that can't keep enough performance when higher resolutions are used (especially via Desktop Fullscreen mode).

Closes #842

Note: Right now here is no performance measure implemented yet. Instead, there are compile-time target architecture macros yet that sets real full-screen mode by default when it's one of 32-bit architectures.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1046,6 +1046,10 @@ else()
         src/core/power/power_sdl.cpp
     )
 
+    if(NOT THEXTECH_FORCE_FULLSCREEN AND NOT EMSCRIPTEN)
+        add_definitions(-DRENDER_FULLSCREEN_TYPES_SUPPORTED)
+    endif()
+
     if(NINTENDO_WIIU)
         list(APPEND THEXTECH_SRC
             src/core/wiiu/msgbox_wiiu.cpp
--- a/changelog.txt
+++ b/changelog.txt
@@ -20,6 +20,7 @@ Changelog for 1.3.7.1-dev
 -3DS: fix bug where gameplay could slow down when audio performance is reduced (@ds-sloth)
 -Implemented support for Power-of-Two only textures on some platforms via SDL Render (@Wohlstand)
 -Adjusted the frame-skip behaviour to address the performance slowdown on some systems (@Wohlstand)
+-Added exclusive full-screen mode support to maintain good performance on weak devices (@Wohlstand)
 
 Changelog for 1.3.7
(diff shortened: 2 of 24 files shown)


ds-sloth:
Looks mostly good. I'll actually test it soon. Some quick tasks:

- [ ] Update strings (eg "real" should be "exclusive", this will be easier for users to understand). I can help with this.
- [ ] Add strings to translation templates
- [ ] Make sure the setting is not available on Emscripten
- [ ] Use better heuristic in OpenGL mode: desktop if OpenGL framebuffer is available, and exclusive otherwise
- [ ] Optional: try to obtain a larger standard resolution if the requested resolution is unavailable (eg, use VGA if SNES fails)

Wohlstand:
>  Update strings (eg "real" should be "exclusive", this will be easier for users to understand). I can help with this.

Ye, feel free, but first let CI complete its build first.

> * Optional: try to obtain a larger standard resolution if the requested resolution is unavailable (eg, use VGA if SNES fails)

This SDL2 already does, it even has a function to obrain the "nearest mode" from the standard list.

>  Add strings to translation templates

I will add them once I will test/polish the thing on my machines

>  Use better heuristic in OpenGL mode: desktop if OpenGL framebuffer is available, and exclusive otherwise

A good note

Wohlstand:
Okay, looks like it's done now

ds-sloth (approved):
I don't think this solution is fully ideal (eg it doesn't seem to work on Wayland) but it's better than not implementing the feature at all. We can polish it more later.

Please merge this to main in three steps:
(1) SDL2 POT support 
(2) Frameskip changes -- and file issue for improving the frameskip logic more
(3) Physical fullscreen support

Wohlstand:
Done, please make a look!

ds-sloth:
This should be fine to merge now -- I think that doing so with the rebase method will be best.

PR #938, opened 2025-04-04
Improve the mechanism of the exclusive full-screen working (#920)

### LINKED ISSUE #920: Stabilize the exclusive Fullscreen logics
Stabilise the logic of desktop/exclusive full-screen modes and switching (a new feature to give the better support for the legacy PCs), right now here is a pile of logical mistakes that still needs an attention.

As current method of combining with Desktop fullscreen system works very unstable, and it's need to do next:
- Add separated "screen mode" advanced option that will allow select one of multiple exclusive fullscreen resolutions.
- Resolutions in the Video section will only toggle in-game resolution, neither switching the exclusive video mode.

Wohlstand:
I took the effort to reorganise the whole mechanics of exclusive full-screen system to make it work stable. Rather than syncing in-game resolution with exclusive full-screen, I made it being a separated setting that can be configured. In addition, I added the configure of colour depth (You can toggle between 16-bit and 32-bit mode, the 16-bit mode is primarily important on video cards that can't render normally 32-bit canvases. Visually it's almost no difference).

**Important note:** on S3 Savage 4 the switching between full-screen and windowed modes, also toggling resolution and colour depth causes a crash (possibly a bug of the driver bundled with Windows XP SP0 where I tested everything, but I am not fully sure).

Fixes #920

Merge by squish!


ds-sloth:
I read all of the code and it looks good. I also tested it on Linux under GNOME Wayland (with both the X11 and Wayland SDL video drivers).

Under Wayland it seems like nothing changes in exclusive mode, but X11 (XWayland) emulates exclusive fullscreen very nicely with stretch-to-fill.

I'm happy for this to be merged now.

ds-sloth (approved):

Wohlstand:
Super!

PR #1077, opened 2025-12-22
Initial work on PSP support and other thingies

Wohlstand:
Here I made an initial work that implements the support for the Sony PSP console. Additionally this implements the "half-pixel" mode feature for the SDL renderer (and OpenGL as well) that forces entire graphics to be drawn with genuine single-pixel, so, every 2x2 will be drawn as a true pixel. And GIF records and screenshots will be also 2x smaller.

## Known PSP issues:
- Sound works, BUT, limit of 60 file descriptors really hurts, I can't open more than 60 music streams without making entire thing to fail. And therefore, to play the thing properly, it's required to disable the sound, just.
- Some textures fails to load.
- Small effects gets rendered incorrectly. On desktop using new-made half-pixel mode just works, no glitches.

## Known Half-Pixel mode
They MUST be fixed before I will split this into separated pull-request
- [x] On OpenGL renderer, the screenshots and recording capture produces distorted frame. It's expected to take the half-sized texture on the query end. But instead it returns a damaged texture (example below).
- [x] Custom scale factors on SDL render should work correctly (right now they are confused).

Scr_2026-01-02_05-43-28

# TODO list
- [ ] Finish the proper screen scale support for platforms where frame buffer is absent such as legacy PC video cards, Mac Mini G4, and some other things.
- [ ] Upgrade the mbediso to allow it support shared file descriptor per thread to avoid it produce over 60 file descriptors on PSP, otherwise audio support is almost impossible without dirty tricks.
- [ ] Fix the problem of small effects sprites becoming dirty while rendered.
- [ ] Ensure the image scaling algorithm is able to schrink large textures without exploding the RAM. Alternatively, suggest the use of pre-shrinked textures with the special comment of the real size.


Wohlstand:
After sleep, first I going to fix all build bugs I "implemented" during this process. Later I will try to split entire thing into half-pixel mode support as a feature for the everything, and the PSP platform support itself.

Wohlstand:
Just now I inpected all build fails and applied some tweaks to fix the build. Some of these build errors are dumb.

Wohlstand:
Just now I rebased everything, so, it should be on top of `main` and no mixture of merge commits. I made full sure that diff of rebased version is equal to the original messy version.

Wohlstand:
What about the half-pixel part, I have an idea - I will try to copy commits related to the half-pixel mode to the new branch, then let it being merged first, and then I will rebase PSP over that, having these commits being cleaned up from THIS branch. Additionally I want to try backport the half-pixel mode and PSP support to the stable branch too. The half-pixel mode literally a mandatory thing, otherwise visible area on PSP is too small if it gets double-sized.

ds-sloth:
The changes above are all of my suggested changes. They include a few review notes with FIXMEs. Please take a look and confirm whether everything still works well on your PSP as well as the other targets.

One thing to note is that PSP-specific logic should be relatively confined. It's not good to have too much PSP logic scattered throughout the whole engine. For instance, instead of PSP-specific logic in `change_res.cpp`, we should make the renderer reset TargetW/TargetH if the desired size of framebuffer cannot be created (this is safe because TargetW/TargetH are not used by game logic).

I think it's a very good idea to first port half-pixel, then port the rest. If you can make a new PR for half-pixel I'll be very happy to review it. Half-pixel should also be possible to backport to main, but I'm very confident that PSP support will not work well without XTA support.

Can you tell me whether thread-local storage exists on PSP? If it does, it should be possible to implement per-thread file-handle pools in mbediso which will fix the max open files issue.

Wohlstand:
Just a note for me: the GIF recording is failed when running recording of half-pixel with OpenGL render:

[Scr_2026-01-02_04-31-07.webm](https://github.com/user-attachments/assets/78593fc8-ba43-4720-bb17-b7e659f3ca55)

But it works as charm on SDL2 render:

[Scr_2026-01-02_04-30-59.webm](https://github.com/user-attachments/assets/cb5919e1-01bd-49d8-9bdb-63e7ce2657f6)

Wohlstand:
> Can you tell me whether thread-local storage exists on PSP? If it does, it should be possible to implement per-thread file-handle pools in mbediso which will fix the max open files issue.

I am not sure about that. However, I do have an idea to make any "music"-based sounds to play just by opening-closing way without keeping permanent file descriptor even that migth cause a kind of hangs on every sound play trigger as I guess. And, that can only be tested if try to make a simple demo that produces such file descriptors. The demo I made already tests only local file descriptors.

Wohlstand:
Tested just now, and unfortunately:
- Feels like half-pixel does NOT works as expected, the dynamic resolution is totally wrong! (it just went to small screen, but it should use PSP's native resolution with the half-pixel mode enforced)
- Larger resolutions still doesn't work, totally. So, a fresh evidence made just now:

https://github.com/user-attachments/assets/5af4ac13-6296-48f0-9274-07e0e121f34c

src/change_res.cpp
```diff
@@ -126,15 +126,6 @@ void UpdateInternalRes()
...
 
-#if defined(__PSP__)
-    // Limitation of texture size for the frame buffer (de-facto 512x512, with half-pixel mode is 1024x1024)
-    if(req_w > 1024)
-        req_w = 1024;
-
-    if(req_h > 1024)
-        req_h = 1024;
-#endif
-
```
Wohlstand: Unfortunately, it is NOT enough to just limit the TargetW/TargetH size, the more complex logic just gets ruined around and outside the renderer.

Wohlstand:
ACTUALLY! The problem that new_ScreenW / new_ScreenH are LARGER than TargetW / TargetH, and therefore, result doesn't render.

Wohlstand:
Also! I got a think where the half-pixel can be useful too - there are such portable consoles on Linux-based firmware and if I am not wrong, they might have small screen, and the half-pixel mode should really toggle the screen size (if it was too small, then toggle it 2x bigger, something like that can be done on Portmaster build since it relies on physical screen size rather than on dynamic window size).

Wohlstand:
I rebased branch again (that means I done ForcePush)

Wolf3s:
Hey there can i use this branch for porting to PS2, since it has some optimized fixes from it?

Wohlstand:
Hello!

You might want to contribute PS2 port to this branch directly, there are still some imperfections that I should solve before I will merge this. Otherwise wait once I will finish this, and then apply PS2 support after that. And also, would you like to share the PS2 toolchain that builds everything? I want to insert it into the docker container that I already use for homebrew builds, so, it will be used to produce automated builds in the next time.

Wohlstand:
And ye, about PSP, I do have real hardware to test, so, I doing this port.

Wohlstand:
Also! I want to note that sometimes I rebase this branch, and it might become conflictish, so, be aware to rebase your local changes once you find my branch was been rebased.

Wohlstand:
@ds-sloth , just in a case I added the TODO list into the main post, please tweak it if needed (maybe I forgot to apply some?).

Wolf3s:
> Also! I want to note that sometimes I rebase this branch, and it might become conflictish, so, be aware to rebase your local changes once you find my branch was been rebased.

I will wait to this being merged first, but my intentiton was use this port to study at least the engine to port to the ps2.

Wolf3s:
> And ye, about PSP, I do have real hardware to test, so, I doing this port.

Ok, i have also a PSP to test it if i want myself to compile it.

Wohlstand:
> > Also! I want to note that sometimes I rebase this branch, and it might become conflictish, so, be aware to rebase your local changes once you find my branch was been rebased.
> 
> I will wait to this being merged first, but my intentiton was use this port to study at least the engine to port to the ps2.

That's should be fine then :)
And if you dare to contribute the final thing, that's will be very appreciated! But don't worry, if something wrong, me and my co-developer @ds-sloth will note you about mistakes and you can learn that. :wink:

PR #1153, opened 2026-06-26
Add disable-fullscreen-toggle config option

bvandenbogaard:
Allows the runtime fullscreen toggle (~~F7 hotkey~~  ALT+ENTER) to be disabled via configuration. I run TheXTech on a personal arcade cabinet where the keys ALT and ENTER are buttons that the controller board sends out. When playing two player mode, those keys get toggled in combination really quickly and cannot be remapped, causing the toggle to be accidentally triggered.

Config key: disable-fullscreen-toggle = true
Section: [video-system]

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -190,8 +190,13 @@ public:
     opt<bool> fullscreen{this, defaults(false), {}, Scope::Config,
         "fullscreen", "Fullscreen", nullptr,
         config_fullscreen_set};
+
+    opt<bool> disable_fullscreen_toggle{this, defaults(false), {}, Scope::Config,
+        "disable-fullscreen-toggle", "Disable fullscreen toggle", nullptr,
+        nullptr};
 #else
     static constexpr bool fullscreen = true;
+    static constexpr bool disable_fullscreen_toggle = true;
 #endif
 
     opt_enum<std::pair<int, int>> internal_res{this,
--- a/src/graphics.cpp
+++ b/src/graphics.cpp
@@ -546,6 +546,9 @@ static inline int s_round2int(num_t d)
 void ChangeScreen()
 {
 #ifndef RENDER_FULLSCREEN_ALWAYS
+    if(g_config.disable_fullscreen_toggle)
+        return;
+
     // shouldn't be possible
     if(g_config.fullscreen.m_set > ConfigSetLevel::user_config)
     {


ds-sloth:
How about going to your controls profile and unsetting the hotkey?

Controls -> Keyboard -> Keyboard 1 -> Hotkeys -> Fullscreen -- hit the Alt Run key to unbind the default hotkey.

bvandenbogaard:
My apologies, I meant ALT+ENTER instead of F7. I have unbound the fullscreen toggle in my control profiles, but ALT+ENTER is fixed in code and is not configurable.

ds-sloth:
Oh, interesting! That's not a case I would have imagined -- F7 I could expect, but not Alt. I might suggest trying to reconfigure the system not to use Alt for a gameplay input. Alt is treated as a modifier key and could disrupt other games / applications as well.

Do you build from source? If so, it may be reasonable for you to set `-DTHEXTECH_FORCE_FULLSCREEN=On` when invoking `cmake`. If that isn't satisfactory, let us know, and I'll add an option to the keyboard settings that disables all default key bindings.

Wohlstand:
Actually ye, if it's need to run an application being always-full-screen, it's need to build it with `-DTHEXTECH_FORCE_FULLSCREEN=ON`, primarily it's used for some mobile platforms that run a desktop-based operating system (primarily for PortMaster target that is based on desktop Linux, but targeted to run on portable consoles).

ChristianSilvermoon:
I will say there are other games in which Alt is a default Keybind, including Helldivers II (which uses it for diving)

But other than that I don't know what to suggest beyond allowing hotkeys to be rebound.

Wohlstand:
@ds-sloth, shall we ever merge this? Or, as I think, implement something different like the "kiosk mode" where app always starts as full-screen without the windowed mode support. That's to allow users to fine tune the pre-build installation without the necessary to rebuild from the source code. And it's fine to keep this option being text-only, without binding into the menu sincee it's a kind of "extra advanced features". Or, bind it into "advanced" menu in any way?

ds-sloth:
I don't think so, no. I think it's better to just use the build flag. This is a very advanced case, and I don't think it's worth our maintenance burden.

Wohlstand:
The build flag is suitable for users who CAN understand how to build projects from the code, ye. But what about dummies who can't? Shall we release a book "Building TheXTech for dummies"? (Just kidding xD)

Wohlstand:
Anyway, understanding the fact that this is not the mass request, then it's better to just guide users to build project with enabling the "force full-screen" mode done.

PR #1155, opened 2026-07-03
Fix compilation on newer MSVC versions

Tkain:
Trying to compile TheXTech using newer versions of MSVC (tested with the latest MSVC as of writing) leads to a build failure because of fmt's usage of `stdext::checked_array_iterator`, which was [deprecated in 2023](https://github.com/microsoft/STL/pull/3818) and [removed in 2025](https://github.com/microsoft/STL/pull/5817).

This PR fixes this by disabling the `FMT_SECURE_SCL` macro for versions of MSVC in which this class has been deprecated or removed.

### DIFF
--- a/lib/fmt/fmt_format.h
+++ b/lib/fmt/fmt_format.h
@@ -59,7 +59,7 @@
 # define FMT_HAS_STRING_VIEW 0
 #endif
 
-#if defined _SECURE_SCL && _SECURE_SCL
+#if defined _SECURE_SCL && _SECURE_SCL && _MSC_VER < 1938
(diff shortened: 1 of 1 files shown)


Wohlstand:
Hello! Please also make a check of `defined(_MSC_VER)` or something to ensure that build will work on other compilers.

Wohlstand:
Actually, after this quick fix I will need to check for the update of mainstream fmt thing and re-appply my tweaks again. And even better, attempt to push these things to mainstream to don't bother about them in the future.

Wohlstand:
Huh, I forgot to take this!

PR #1157, opened 2026-07-10
Support for iOS and tvOS

Wohlstand:
This is still a work-in-process thing, but right now, game boots and works! There are few minor thingies remained until this gets ready for merge. And before that, this draft will serve as a way to easily review the summary of all done tweaks.

IMG_0055

**Thingies are remained:**
- [x] Adjust the render blocking, probably just disable this, but make few more tests to make sure it won't fail while switching between apps, etc.
- [x] Implement offsets for controls. On iPhone X and similar devices it's pretty inconvenient to use buttons that gets placed too close to edge. This feature will be useful for other similar phones including Android-based, as buttons gets placed inconveniently and it's need to adjust the offset.
  - [x] Make automatic defaults depending on device model that has or don't have the Face ID cut presence, or do they have too thin edges at top and bottom sides of the screen so that makes inconvenient for fingers such placement.
- [x] Fix the vibration support, for now it doesn't works, it's need to utilise the [CoreHaptics](https://developer.apple.com/documentation/corehaptics?language=objc) for iOS 13+, and try out workarounds to call built-in vibration commands on older iOS.
- [x] Implement detection of the physical size to set the default sizes for controls.
- [x] Build and test the thing on iPhone 4s with the iOS 9.
- [ ] Try to find a better way to finalise the application as it's a typically bad practice to call the `exit()` function as it leads to some internal troubles on the iOS. Right now, the programmical calling of the home button is done together with the two seconds waiting until calling the `exit()` to finish the thing.
- [x] Make a CI script to build the project automatically.
  - [x] Build modern builds (arm64, iOS 12+) 
  - [x] Build tvOS builds (arm64, tvOS 12+)
  - [x] Build legacy builds (armv7,armv7s, iOS 9+)
- [x] For the tvOS, adjust the workflow with paths:
  - [x] Game packs must be found also at Resources in similar manner as on macOS, but allow multiple, and worlds too!
  - [x] Save gamesaves at Library/Preferences directory and don't write anything also to Documents (in Real hardware it's impossible to save anything here, only Cache is writeable, it's a limitation of tvOS itself!). It's possible only using [`NSUserDefaults`](https://developer.apple.com/library/archive/documentation/LegacyTechnologies/WebObjects/WebObjects_3.5/Reference/Frameworks/ObjC/Foundation/Classes/NSUserDefaults/Description.html#//apple_ref/occ/cl/NSUserDefaults) interface.
  - [x] Implement ~~XZ~~ ZIP compress/decompress to save the space for gamesaves (only 500 KB allowed to be preserved, if too much gamesaves and large settings, that will waste the space very quickly!).

- [ ] After the thing will be completed, make the maunal on how to deal with the thing including cases that requiring to reinstall the app every two days and other cases. And of course, merge the thing into **main**.

## Some known issues:
- [x] Random crash happens during loading time, on attempt to destroy some texture: `-[IOGPUMetalCommandBuffer validate], line 216: error 'commit command buffer with uncommitted encoder'`
- [x] Sometimes the infinite vibration might happen, phone starts to buzz forever. The only work around is turn app into background, wait some, and open it back.
- ⛔ V-Sync is enforced and no way to disable it. Just. Seems it's better to keep it always-on and just hide the ability to toggle it, on iOS it's impossible.


Wohlstand:
What about the vibration support, it's a problem on iOS as I investigated: the API provides only very limited way to trigger vibration without controlling of duration and other things that makes Haptic thing being impossible here. However, such vibrations might be used for bombs and hurts at least. It's need to test that. Otherwise just completely disable that and allow only Bluetooth / Wired gamepads work with the Haptic thing.

Wohlstand:
Implemented DPI handling, glad SDL2 helped with some function a lot, otherwise iOS sucks with lack of physical stats reporting.

Wohlstand:
Built for iPhone 4S with iOS 9 and game works successfully!

IMG_0025

IMG_0023

Wohlstand:
What about haptic support, it's possible to investigate for this thing:

https://developer.apple.com/documentation/corehaptics?language=objc 

And if it will work, then attempt to implement at the SDL2 with a note that entire thing will work for iOS 13 only. And on iPhone 8+ only.

Wohlstand:
Offsets for controls implemented as user options, but right now it's not set the automatical default depending on device type (to detect presence of the Face ID area cut and something).

Wohlstand:
I think, the basic support is fully completed, and remained tasks can be turned into task lists to finish just in the mainstream.

Wohlstand:
Build for the legacy iOS is done perfectly! And I think, I will switch modern builds from Xcode to the Ninja build as it goes MUCH FASTER than done via Xcode, and make sure done builds are equal to these are done by Xcode.

Wohlstand:
Vibration support has been fully implemented, but not yet tested on iOS 9. The remained problem with Haptics Engine is that render shortly hangs every tap. I am not yet sure why that happens, seeking forums around...

Wohlstand:
I made a fix that makes detection of curved screens and assings the overscan, so, on rectangular screens it works as usually, but on curved screens it gets adjusted.

Wohlstand:
For the tvOS I implemented support for two directories at the IPA file:
- "assets" to store game packs
- "worlds" to store episodes per game pack (they should be placed at sub-directory like "worlds/smbx/Super World/")

And then, custom worlds gets detected:
Снимок экрана — 2026-07-20 в 01 03 22

Game saves, logs, etc. for now gets stored at caches. Settngs and gamesaves must be written using the [`NSUserDefaults`](https://developer.apple.com/library/archive/documentation/LegacyTechnologies/WebObjects/WebObjects_3.5/Reference/Frameworks/ObjC/Foundation/Classes/NSUserDefaults/Description.html#//apple_ref/occ/cl/NSUserDefaults) thing.

Wohlstand:
During recent times, I made:
- Handling of UIScene thing, so, no more boot screaming about any future crashes. I made the backporting of the feature from the SDL3 to the SDL2 at AudioCodecs.
- Verified work on the real tvOS hardware.
- Fixed the game selection menu being immediately closed after being open from the main menu.
- Fixed some glitches spammed into terminal during background work.

The thing looks more and more completed now! The last real thing is left the saving of settings at the NSUserDefaults. I think I will implement the thing that works close to the Emscripten's one using the same method. But instead, I will dump any significatn data into archive and save into the settings area that will survive the crash nuking on tvOS.

Wohlstand:
Okay, I implemented the full-featured preservation ofsettings and game saves at the NSUserDefaults being packed as a ZIP archive. These settings gets restored back once `settings` directory is absent during the boot.

Wohlstand:
@ds-sloth , now thing is really ready for review. Please review the thing once you can, and then I will backport it to the `main` after merging to here.

src/core/sdl/events_sdl.cpp
```diff
@@ -35,14 +35,42 @@
...
+    // case SDL_WINDOWEVENT:
+    //     SDL_memcpy(&self->m_event, event, sizeof(SDL_Event));
+    //     self->processEvent();
+    //     return 0;
+
+    // default:
+    //     return 1;
+    // }
+}
+#endif
```
Wohlstand, line 64: Whoops, I think I should clean this up.

Wohlstand: Done.

src/core/render.h
```diff
@@ -58,7 +58,7 @@ constexpr int TargetOverscanX = 28;
 
 #elif defined(THEXTECH_IOS)
 
-extern int TargetCameraOverscanX;
+constexpr int TargetCameraOverscanX = 0;
```
Wohlstand, line 61: I am a bit confused and I just wanted to note that the overscan size on iOS is different between devices: newer devices with round corners, and old devices and iPhone SE devices with sharp corners.

Wohlstand:
By the way, @ds-sloth, if you know how to make OpenGL ES on iOS quicker, then feel free to apply the tweak. I attempted quickly, and result started with a black screen. I didn't yet tested on iPhone 4S and iOS 9 to see, does the same happens or not. And if really the OpenGL ES no longer works on iOS, then it's need to implement direct use of the Metal API for the accurate render. At least it's possible to debug that on the macOS until attempting to run on iOS. But, as this might take a while, then it should go as a separate task and iOS will just live with the SDL render.

ds-sloth:
I cherry-picked all of the safe changes to main and stable, then rebased this branch. I added three tiny commits myself, and here are my comments about the remaining changes:
* Regarding the touchscreen graphics threading errors: this should be fixed by reloading the touchscreen graphics in the main thread, not by (1) unloading them on the main thread and then (2) not drawing the touchscreen controls during a reload from the asset pack screen. The touchscreen controls are needed to keep a smooth appearance during the asset pack screen -> main menu transition.
* I would personally strongly recommend using a screen boundary which doesn't intersect the camera zone ("notch"), instead of filling the screen and using TargetOverscanX as a workaround.
* The added check `FontManager::isInitied()` in `render_sdl.cpp` is unnecessary (given the error checking of `FontManager::printText`), and it makes the source code harder to read.
* Make sure to carefully test the changes to `frmMain::restartRenderer` and confirm that they are necessary (to switch between SDL2 and SW renderers, I assume?).
* `handle_ios_events` may lead to some subsystems being accessed after they are partially destroyed. If you partially deinit the system here, you may want to call `exit()` as well. Otherwise, I'd advise removing the deinitialization calls.

src/core/render.h
ds-sloth: Overscan size is different, but *Camera* overscan is the same. Here's the comment from that commit:

```
iOS: keep camera overscan as a constant

(This is a unique hack used only on the 3DS, and it doesn't work without parallax layers.)
```

ds-sloth:
> OpenGL ES

As far as I can tell, GLES should still work, so I'm not sure what the issue is. It's also possible that ANGLE will "just work" here. It has a Metal backend upstream but I'm not certain whether it supported Metal at the time we forked our C++11 version. You could try to build it with ANGLE and see what happens.

Wohlstand:
> The added check FontManager::isInitied() in render_sdl.cpp is unnecessary (given the error checking of FontManager::printText), and it makes the source code harder to read.

It IS needed, because it CRASHED the app in my test case with the storage overflown warning as it gets printed always even during reboot of game pack that causes textures and fonts being termpoarily unloaded. Otherwise it's need to improve the presence of the loaded fonts and textures and just refuse to print until necessary textures gets finally loaded.

ds-sloth:
I'll test this. The `SuperPrint` family is never hot, so any error checking should occur inside the printing functions instead of outside. (And this will make the engine safer in general.)

Wohlstand:
> Regarding the touchscreen graphics threading errors: this should be fixed by reloading the touchscreen graphics in the main thread, not by (1) unloading them on the main thread and then (2) not drawing the touchscreen controls during a reload from the asset pack screen. The touchscreen controls are needed to keep a smooth appearance during the asset pack screen -> main menu transition.

I made what I made as it goes simpler and does not overcomplicates everything. If implement a kind of the router to order the main thread to load a texture during loop, it would be a serious overcomplication, and I wanted to avoid it.

Wohlstand:
> I would personally strongly recommend using a screen boundary which doesn't intersect the camera zone ("notch"), instead of filling the screen and using TargetOverscanX as a workaround.

I think to make some a bit later. But, I have a separate idea that would be better generally for any touch-screen devices, not just for such iOS devices: allow section scroll farther than the edge of screen so keep player being close to centre, and that will also solve a problem of fingers covering areas where player is.

Wohlstand:
> handle_ios_events may lead to some subsystems being accessed after they are partially destroyed. If you partially deinit the system here, you may want to call exit() as well. Otherwise, I'd advise removing the deinitialization calls.

Does it's allowed to call an exit() here? Actually this deinitialisation I added primarily to gracefully close logs and other things when application was terminated by swyping from a list of background apps.

Wohlstand:
> Make sure to carefully test the changes to frmMain::restartRenderer and confirm that they are necessary (to switch between SDL2 and SW renderers, I assume?).

I need to test this, ye.

ds-sloth:
> I made what I made as it goes simpler and does not overcomplicates everything. If implement a kind of the router to order the main thread to load a texture during loop, it would be a serious overcomplication, and I wanted to avoid it.

Here's a commit showing my idea -- feel free to revert it if you don't like it.

> `textInitied` check

I also added a commit removing this and making the `textSize` call safe when the font subsystem isn't initialized. I think that's a good idea because it reduces the crash surface of the program.

> Does it's allowed to call an exit() here? Actually this deinitialisation I added primarily to gracefully close logs and other things when application was terminated by swyping from a list of background apps.

I'm not certain about this. I don't know much about the iOS internals or what the callstack looks like here (though I thought it was a normal SDL event poll?).

Wohlstand:
1) First I should make full sure it will don't crash. I'll test the thing on my devices and see will it survive or not.  The goal is to not let it crash.
2) :+1:
3) I got that from the manual document from the SDL2 repo about iOS thing, it noted about events are should be handled in this way before they can be routed into normal SDL events chain.

ds-sloth:
I think for (3), let's just test this extensively. Like how 3DS Home Menu exit is dangerous, this is likely a dangerous area that is better to test than to logically reason about.

Anyway, I'll review some more. It'd be nice to merge this when possible. Can you share a link to your instructions for how to test on hardware?

lib/AppPath/private/app_path.cpp
```diff
@@ -155,6 +155,7 @@ std::vector<std::pair<std::string, AssetsPathType>> AppPathManager::assetsSearch
 
 #ifdef APP_PATH_HAS_EXTRA_WORLDS
         // 3DS: add assets from additional romfs packages
+        // tvOS: add extra episodes per assets from the bundle package
```
ds-sloth, line 158: Add comment: "(the list has one item on tvOS)"

lib/AppPath/private/app_path_ios.mm
```diff
@@ -0,0 +1,135 @@
...
+}
+
+void AppPathP::syncFS()
+{
+    /* Run the FS synchronization (Implement this for Emscripten only) */
+}
+
+std::string AppPathManager::bundleResourcesPath()
+{
+    return s_bundlePath;
```
ds-sloth, line 134: This is ugly but I understand why it got added.

src/core/sdl/events_sdl.h
```diff
@@ -33,6 +33,10 @@ class EventsSDL final : public AbstractEvents_t
     bool m_gotExit = false;
 #endif
 
+#if defined(THEXTECH_IOS) || defined(THEXTECH_TVOS)
+    static int handle_ios_events(void *userdata, SDL_Event *event);
```
ds-sloth, line 37: Maybe make this a static non-member function inside events_sdl.cpp instead of a static member function. This looks like it's exposing implementation details.

ds-sloth (approved):
I didn't review the AppPath, Logger, CMake, or CI changes closely. My expectation is that any bugs in these areas will become clear with testing.

Is miniz really necessary? Are we not able to use zlib or lz4 here?

Why are there multiple copies of some of the new icons, with filenames like `_1.png`?

Wohlstand:
1) :+1: 
2) If zlib is able to pack a full featured ZIP archive, it could work. If it's only an algorithm, then a kind of the packer will be required. MiniZ is a ZIP archive packer that I use to pack entire directory, not just one single file.
3) This seems a result of macOS how it packs ZIP archives or something.

lib/AppPath/private/app_path.cpp
Wohlstand: There are many directories, it's possible to bundle multiple games with one package, and during tests I did that. So, inside one game there are multiple games. Like, user can inject multiple game packs and custom episodes into IPA and install to the tvOS and that's a guarantee the content will survive the nuke of cache.

lib/AppPath/private/app_path_ios.mm
Wohlstand: It's required to get bundled resources especially fallback images for touch controller.

src/core/sdl/events_sdl.h
Wohlstand: The primary reason that I made it member so it can access private fields from the pointer to self referred as `userdata`.

Wohlstand: If make a non-member, then referring as a friend. Still the reference appears at the header. And I better just keep as is.

ds-sloth: > so it can access private fields

I see, yes, this makes good sense.

lib/AppPath/private/app_path.cpp
ds-sloth: As long as you've tested this, I won't complain. But do note that it might be possible to implement this as a single `Multiple` root instead of multiple `Single` roots.

ds-sloth:
> miniz

I see, I didn't realize this was doing a full directory pack. Does `miniz` actually pack and unpack standard ZIPs? If so, this could also be extremely helpful for our Emscripten import/export functionality.

> icons

So, can we remove all of the extra icons? It's always good to keep the repo size as small as possible.

Wohlstand:
> So, can we remove all of the extra icons? It's always good to keep the repo size as small as possible.

Oh wait, repo? They are all required to comply with different sizes as Xcode's grid referred me, and I filled everything. Different devices, places, etc. And I made icon of different sizes and metrics.

PR #1178, opened 2026-09-04
Single-file saves

ds-sloth:
WARNING:
(1) This PR affects saving. Please be cautious and backup any saves before trying it.
(2) This PR introduces a new way of tracking gameplay time and fails. Post-PR TheXTech can read timer/fail records from pre-PR TheXTech, but pre-PR TheXTech cannot read records from post-PR TheXTech. This means that your timer/fail data will be lost if you resume a post-PR gamesave on a release build.

@Wohlstand, we should probably cherry-pick this to stable as well, and then merge this to all branches immediately before the next stable release.

PR features:
* Gamesaves now use a single SAVX file, without additional timer and fails files.
* The SAVX file is updated (without a full save) any time that the timer or fails files would have been updated, by appending new entries to the SAVX file.
* This speeds up the save/load processes, and makes it easier to track and transfer saves.


Wohlstand:
Delayed. I'll merge this after the 1.7.4 stable release.

